PR #9, opened 2021-05-22
Add actx.np.reshape

thomasgibson:
This PR adds `reshape` to the array context numpy namespace for pyopencl arrays. Needed for https://github.com/inducer/meshmode/pull/192


inducer:
Thanks! Refactored a bit in 71dfd32.

PR #13, opened 2021-05-24
Add to_numpy and from_numpy on containers

alexfikl:
Ideally this would just be a `rec_map_array_container(actx.from_numpy, ary)`, but the serialization gets confused by `ndarray`s that aren't object arrays.

I'm not quite sure how to fix that besides making `is_array_container` singledispatched again.

(NOTE: this was mostly inspired by trying to remove [a hack in pytential](https://github.com/inducer/pytential/blob/2f8b631a75925aabad080ace5d0c4d68d4e9ded3/pytential/symbolic/matrix.py#L229))

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -15,6 +15,11 @@ Freezing and thawing
 ~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: freeze
 .. autofunction:: thaw
+
+Numpy conversion
+~~~~~~~~~~~~~~~~
+.. autofunction:: from_numpy
+.. autofunction:: to_numpy
 """
 
 __copyright__ = """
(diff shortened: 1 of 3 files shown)


arraycontext/container/traversal.py
```diff
@@ -265,4 +273,34 @@ def thaw(ary, actx):
 
 # }}}
 
+
+# {{{ numpy conversion
+
+def from_numpy(actx, ary):
```
alexfikl, line 279: In case this gets in, `from_numpy(actx, ary)` or `from_numpy(ary, actx)`?

inducer: Yes, let's go for consistency with `thaw` and `freeze`.

alexfikl: Updated in bd76184.

inducer:
Thanks!

PR #14, opened 2021-05-26
move pytato array context from meshmode

matthiasdiener:
Originally at https://github.com/inducer/meshmode/pull/116


arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,371 @@
+"""
+.. currentmodule:: arraycontext
```
inducer, line 2: This should have some basic tests in place. Quite possibly, the existing tests for the pyopencl namespace should be broadened to include this array context. This could probably be done by making a new, modified/parametrized (?) version of the `actx_factory` fixture.

matthiasdiener: Modified the test generator to also run the same tests for the PytatoArrayContext. Is this what you had in mind (minus the function argument)?

matthiasdiener: Current test output:

```
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-sin-1]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-exp-1]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-arctan2-2]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-minimum-2]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-maximum-2]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-where-3]
FAILED test/test_arraycontext.py::test_array_context_np_workalike[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-conj-1]
FAILED test/test_arraycontext.py::test_array_context_np_like[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ones_like-1]
FAILED test/test_arraycontext.py::test_actx_stack[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>]
FAILED test/test_arraycontext.py::test_dof_array_arithmetic_same_as_numpy[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>]
FAILED test/test_arraycontext.py::test_dof_array_reductions_same_as_numpy[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_manipulation[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ij->ij]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_manipulation[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ij->ji]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_manipulation[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ii->i]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_matmatprods[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ij,ij->ij]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_matmatprods[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ij,ji->ij]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_matmatprods[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-ij,kj->ik]
FAILED test/test_arraycontext.py::test_array_context_einsum_array_tripleprod[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-im,mj,k->ijk]
FAILED test/test_arraycontext.py::test_container_multimap[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>]
FAILED test/test_arraycontext.py::test_container_arithmetic[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>]
FAILED test/test_arraycontext.py::test_container_norm[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-2]
FAILED test/test_arraycontext.py::test_container_norm[<Pytato array context factory for <pyopencl.Device 'pthread' on 'Portable Computing Language'>-inf]
============================== 22 failed, 32 passed, 501 warnings in 8.11s ==============================
```

inducer:
@matthiasdiener Could you please revert a1b97a6?

matthiasdiener:
> @matthiasdiener Could you please revert [a1b97a6](https://github.com/inducer/arraycontext/commit/a1b97a663df5a654e580c4092a901af75a1dff7c)?

sure, done

arraycontext/impl/pytato.py
```diff
@@ -36,11 +36,57 @@
...
+
+        if ary.size == 0:
+            return 0
+
+        if ord == np.inf:
+            return self._array_context.np.max(abs(ary))
+        elif isinstance(ord, Number) and ord > 0:
+            return self._array_context.np.sum(abs(ary)**ord)**(1/ord)
+        else:
+            raise NotImplementedError(f"unsupported value of 'ord': {ord}")
```
inducer, line 89: There's nothing specific to the array context in here. This can move to the base class.

In general, please avoid copying this much code, particularly verbatim like in this case.

matthiasdiener: Refactored in https://github.com/inducer/arraycontext/pull/14/commits/ed2374fda1e00f74bc7c91943f9da1a0da2003f5

requirements.txt
```diff
@@ -5,5 +5,7 @@ git+https://github.com/inducer/pymbolic.git#egg=pymbolic
 git+https://github.com/inducer/pyopencl.git#egg=pyopencl
 git+https://github.com/inducer/islpy.git#egg=islpy
 
-git+https://github.com/inducer/loopy.git#egg=loopy
-git+https://github.com/inducer/pytato.git#egg=pytato
+git+https://github.com/inducer/loopy@kernel_callables_v3-edit2.git#egg=loopy
+git+https://github.com/kaushikcfd/pytato.git@call_loopy#egg=pytato
+
+git+https://github.com/inducer/meshmode.git#egg=meshmode
```
inducer, line 11: This goes the wrong way... what's the rationale for adding this?

matthiasdiener: arraycontext/impl/pytato.py and arraycontext/impl/pyopencl.py use `meshmode.dof_array.DOFArray`

inducer: If you're talking about this:
https://github.com/inducer/arraycontext/blob/23121b12f86faad1ce8cae2556b2f7f19b30fe18/arraycontext/impl/pyopencl.py#L175-L193
That's a deprecated code path that's only there for backward compatibility. It doesn't justify installing anything IMO.

inducer: I don't see any other mention of `meshmode`.

inducer: To provide more perspective: The assertion that this package works without meshmode is more valuable (to me) than the assertion that this busted code path works.

inducer:
> get rid of most of meshmode e9156d5 

:+1: Thanks!

requirements.txt
matthiasdiener: I removed most of the meshmode code in https://github.com/inducer/arraycontext/pull/14/commits/e9156d5dc18e9fb2ca7648e49ecc896221e9fee5 and refactored the DofArray check in https://github.com/inducer/arraycontext/pull/14/commits/ef786c92b715e64a06e74f9b97e83b71700b3b85

matthiasdiener:
I think this is ready for some more deep review @inducer @kaushikcfd . I had to work around some test failures, maybe you have ideas for better fixes.

test/test_arraycontext.py
```diff
@@ -291,8 +295,9 @@ def get_imag(ary):
             (operator.neg, 1, False),
             (operator.abs, 1, False),
 
-            (get_real, 1, False),
-            (get_imag, 1, False),
+            # Not supported in pytato:
+            # (get_real, 1, False),
+            # (get_imag, 1, False),
```
inducer, line 300: Wouldn't this decrease test coverage for the pyopencl actx?

test/test_arraycontext.py
```diff
@@ -249,7 +253,7 @@ def get_real(ary):
         return ary.real
 
     def get_imag(ary):
-        return ary.real
+        return ary.real  # FIXME?
```
inducer, line 256: Explain?

test/test_arraycontext.py
```diff
@@ -307,6 +312,23 @@ def get_imag(ary):
...
+            elif op_func == operator.gt:
+                op_func_actx = actx.np.greater
+            elif op_func == operator.eq:
+                op_func_actx = actx.np.equal
+            elif op_func == operator.ne:
+                op_func_actx = actx.np.not_equal
+            elif op_func == get_real:
+                op_func_actx = actx.np.real
+            elif op_func == get_imag:
+                op_func_actx = actx.imag
```
inducer, line 328: How come pytato doesn't support these as is? Should be straightforward AFAICT...

inducer (commented):
Thanks! Some preliminary comments from a quick scroll.

test/test_arraycontext.py
matthiasdiener: It seemed strange to me that a `get_imag` function returns the real part, but I wasn't sure if this is intentional or not.

doc/array_context.rst
```diff
@@ -12,3 +12,9 @@ Array context based on :mod:`pyopencl.array`
 --------------------------------------------
 
 .. automodule:: arraycontext.impl.pyopencl
+
+
+Array context based on :mod:`pytato`
+--------------------------------------------
```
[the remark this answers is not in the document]
matthiasdiener: Fixed

arraycontext/impl/pyopencl.py
```diff
@@ -391,6 +344,11 @@ def tag_axis(self, iaxis, tags: Union[Sequence[Tag], Tag], array):
     def clone(self):
         return type(self)(self.queue, self.allocator, self._wait_event_queue_length)
 
+    def compile(self, f: Callable[[Any], Any],
```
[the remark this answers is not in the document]
matthiasdiener: Moved the main `compile` definition to the base class and added some docs.

test/test_arraycontext.py
matthiasdiener: Added real/imag support to pytato, restored test.

test/test_arraycontext.py
matthiasdiener: Changed to return the imag part.

kaushikcfd:
> add pytato to dependencies

I think we should follow `pyopencl` and not make pytato a dep.

test/test_arraycontext.py
kaushikcfd: At least the comparison operators: pytato arrays cannot support those.

matthiasdiener:
> > add pytato to dependencies
> 
> I think we should follow `pyopencl` and not make pytato a dep.

Packages using arraycontext now need pytato to run the tests; I'm not sure how to better handle this.

test/test_arraycontext.py
matthiasdiener: removed real(), imag() from this list, left the comparison operators.

arraycontext/loopy.py
```diff
@@ -51,7 +51,6 @@ def make_loopy_program(domains, statements, kernel_data=None,
             statements,
             kernel_data=kernel_data,
             options=_DEFAULT_LOOPY_OPTIONS,
-            default_offset=lp.auto,
```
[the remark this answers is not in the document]
matthiasdiener: With this option set, we get the following error:

```
pyopencl._cl.RuntimeError: clBuildProgram failed: BUILD_PROGRAM_FAILURE - clBuildProgram failed: BUILD_PROGRAM_FAILURE - clBuildProgram failed: BUILD_PROGRAM_FAILURE
Build on <pyopencl.Device 'pthread' on 'Portable Computing Language' at 0x119d64fe0>:
error: /Users/mdiener/.cache/pocl/kcache/tempfile-09-5c-db-e2-bd.cl:16:67: too few arguments to function call, expected 7, have 4

static void lp_nodes(int const nelements, __global double const *__restrict__ nodes, int const nodes_offset, __global double const *__restrict__ resampling_mat, int const resampling_mat_offset, __global double *__restrict__ result, int const result_offset);
__kernel void __attribute__ ((reqd_work_group_size(1, 1, 1))) _pt_kernel(__global double const *__restrict__ _pt_in_0, __global double const *__restrict__ _pt_in, __global double *__restrict__ _pt_out, __global double *__restrict__ _pt_temp, __global double *__restrict__ _pt_temp_0, __global double *__restrict__ _pt_temp_1)
{
  for (int _pt_temp_dim1 = 0; _pt_temp_dim1 <= 2; ++_pt_temp_dim1)
    for (int _pt_temp_dim0 = 0; _pt_temp_dim0 <= 511; ++_pt_temp_dim0)
      _pt_temp[3 * _pt_temp_dim0 + _pt_temp_dim1] = _pt_in_0[3 * _pt_temp_dim0 + _pt_temp_dim1];
  for (int _pt_temp_0_dim1 = 0; _pt_temp_0_dim1 <= 2; ++_pt_temp_0_dim1)
    for (int _pt_temp_0_dim0 = 0; _pt_temp_0_dim0 <= 9; ++_pt_temp_0_dim0)
      _pt_temp_0[3 * _pt_temp_0_dim0 + _pt_temp_0_dim1] = _pt_in[3 * _pt_temp_0_dim0 + _pt_temp_0_dim1];
  lp_nodes(512, &(_pt_temp[0]), &(_pt_temp_0[0]), &(_pt_temp_1[0]));
  for (int _pt_out_dim1 = 0; _pt_out_dim1 <= 9; ++_pt_out_dim1)
    for (int _pt_out_dim0 = 0; _pt_out_dim0 <= 511; ++_pt_out_dim0)
      _pt_out[10 * _pt_out_dim0 + _pt_out_dim1] = _pt_temp_1[10 * _pt_out_dim0 + _pt_out_dim1];
}
static void lp_nodes(int const nelements, __global double const *__restrict__ nodes, int const nodes_offset, __global double const *__restrict__ resampling_mat, int const resampling_mat_offset, __global double *__restrict__ result, int const result_offset)
{
  double acc_j;
  for (int iel = 0; iel <= -1 + nelements; ++iel)
    for (int idof = 0; idof <= 9; ++idof)
    {
      acc_j = 0.0;
      for (int j = 0; j <= 2; ++j)
        acc_j = acc_j + resampling_mat[resampling_mat_offset + 3 * idof + j] * nodes[nodes_offset + 3 * iel + j];
      result[result_offset + 10 * iel + idof] = acc_j;
    }
```

inducer:
This has merge conflicts. Could you take a look?

matthiasdiener:
> This has merge conflicts. Could you take a look?

We'll have to implement ravel now for pytato I guess, but could you please take a look at the rest?

inducer:
> We'll have to implement ravel now for pytato I guess, but could you please take a look at the rest?

That's actually pretty complicated. (@kaushikcfd, correct? How is pytato's `reshape` support?) I would recommend overriding the norm instead. The `ravel` test can be `xfail`ed.

matthiasdiener:
> > We'll have to implement ravel now for pytato I guess, but could you please take a look at the rest?
> 
> That's actually pretty complicated. (@kaushikcfd, correct? How is pytato's `reshape` support?) I would recommend overriding the norm instead. The `ravel` test can be `xfail`ed.

I implemented ravel() (that passes the test), but not sure if it's correct.

kaushikcfd:
Supporting ravel fully (including the order "K" and "A") requires some thought, essentially it asks the question: "What is the order/strides of a pytato array expression's dims?". order = "C" for non-parametric shapes is fully supported till now. order="F" is not difficult to support, but order="K" and "A" are ill-defined for current pytato trunk.

arraycontext/impl/pytato.py
```diff
@@ -148,6 +148,28 @@ def arctan2(self, y, x):
         import pytato as pt
         return rec_multimap_array_container(pt.arctan2, y, x)
 
+    def ravel(self, a, order="C"):
+        def _rec_ravel(a):
+            import pytato as pt
+            if order in "FC":
+                return pt.reshape(a, (-1,), order=order)
+            elif order == "A":
+                if a.flags.f_contiguous:
```
kaushikcfd, line 157: This could be rewritten as just having a fixme and return `pt.reshape(-1, order="C")`.

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,53 @@ class BaseFakeNumpyLinalgNamespace:
...
+            warn("Taking an actx.np.linalg.norm of a DOFArray is deprecated. "
+                    "(DOFArrays use 2D arrays internally, and "
+                    "actx.np.linalg.norm should compute matrix norms of those.) "
+                    "This will stop working in 2022. "
+                    "Use meshmode.dof_array.flat_norm instead.",
+                    DeprecationWarning, stacklevel=2)
+
+            import numpy.linalg as la
+            return la.norm(
+                    [self.norm(_flatten_array(subary), ord=ord)
```
kaushikcfd, line 199: This is pyopencl specific

matthiasdiener: Moved this part of the code to pyopencl.py

arraycontext/impl/pytato.py
matthiasdiener: Done

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,477 @@
...
+        if not isinstance(array, cla.Array):
+            raise TypeError("PytatoArrayContext.thaw expects CL arrays, got "
+                    f"{type(array)}")
+
+        return pt.make_data_wrapper(array.with_queue(self.queue))
+
+    # }}}
+
+    def compile(self, f: Callable[[Any], Any],
+                inputs_like: Tuple[Union[Number, np.ndarray], ...]
```
inducer, line 337: This `inputs_like` thing is a crutch, IMO, and I'm not convinced it's needed. Instead, we could wait until the user calls the "compiled" function and use the data provided in this role. That way, nobody has to make up "fake" data with unclear semantics.

arraycontext/context.py
```diff
@@ -349,6 +351,29 @@ def clone(self):
...
+
+        *f* may be called on placeholder data, to obtain a representation
+        of the computation performed, or it may be called as part of the actual
+        computation, on actual data. If *f* is called on placeholder data,
+        it may be called only once (or a few times).
+
+        :arg f: the function executing the computation.
+        :arg inputs_like: the input arguments to the function.
+        :return: a function with the same signature as *f*.
+        """
```
inducer, line 374: @majosm Here are some docs on how `compile` is supposed to work. I haven't worked spelled out what `inputs_like` is supposed to be because I'm campaigning to get rid of it at https://github.com/inducer/arraycontext/pull/14/files#r651426611.

I'd welcome any feedback you might have.

majosm: Thanks! That answers the questions I had about it. The documentation looks good to me.

arraycontext/impl/pytato.py
kaushikcfd: Thanks, agreed: a7b372a.

matthiasdiener:
I think this is officially ready for review.

.github/workflows/ci.yml
```diff
@@ -117,8 +117,6 @@ jobs:
                 cd "$DOWNSTREAM_PROJECT"
                 echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
 
-                sed -i "/egg=arraycontext/ c git+file://$(readlink -f ..)#egg=arraycontext" requirements.txt
```
inducer, line 120: I think this should stay?

arraycontext/container/traversal.py
```diff
@@ -233,6 +233,76 @@ def wrapper(*args: Any) -> Any:
...
+def rec_keyed_map_array_container(
+        f: Callable[[Tuple[Union[str, int], ...], Any], Any],
+        ary: ArrayContainerT) -> ArrayContainerT:
+    """
+    Works similarly to :func:`rec_map_array_container`, except that *f* also
+    takes in a traversal path to the leaf array. The traversal path argument is
+    passed in as a tuple of identifiers of the arrays traversed before reaching
+    the current array.
+    """
+    return _keyed_map_array_container_impl(f, ary, recursive=True)
```
inducer, line 304: Inline `_keyed_map_array_container_impl`?

arraycontext/context.py
```diff
@@ -349,6 +350,26 @@ def clone(self):
             "setup-only" array context "leaks" into the application.
         """
 
+    def compile(self, f: Callable[[Any], Any]) -> Callable[..., Any]:
```
inducer, line 353: Check the type signature.

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,32 @@ class BaseFakeNumpyLinalgNamespace:
     def __init__(self, array_context):
         self._array_context = array_context
 
+    def norm(self, ary, ord=None):
```
inducer, line 177: I think there should only be one implementation of `norm`, across here and the pyopencl context.

arraycontext/pytest.py
```diff
@@ -31,7 +31,35 @@
...
+        from arraycontext.impl.pytato import PytatoArrayContext
+        return PytatoArrayContext(cl.CommandQueue(ctx))
+
+    def __str__(self):
+        return ("<Pytato array context factory for <pyopencl.Device '%s' on '%s'>"
+                % (self.device.name.strip(),
+                 self.device.platform.name.strip()))
+
+
+def pytest_generate_tests_for_array_contexts(metafunc, use_pytato=True) -> None:
```
inducer, line 62: This should take a list of array context classes.

test/test_arraycontext.py
```diff
@@ -727,9 +745,51 @@ def test_norm_ord_none(actx_factory, ndim):
     norm_a_ref = np.linalg.norm(a, ord=None)
     norm_a = actx.np.linalg.norm(actx.from_numpy(a), ord=None)
 
+    if not np.isscalar(norm_a):
```
inducer, line 748: Both array contexts should handle reduction returns consistently.

.github/workflows/ci.yml
matthiasdiener: Restored

arraycontext/context.py
matthiasdiener: Changed to `Callable[..., Any]`

.github/workflows/ci.yml
```diff
@@ -116,7 +116,8 @@ jobs:
 
                 cd "$DOWNSTREAM_PROJECT"
                 echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
-
+
```
inducer, line 119: Extra whitespace

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,535 @@
...
+        def preprocess_arg(arg):
+            if isinstance(arg, cla.Array):
+                return self.thaw(arg)
+            else:
+                assert isinstance(arg, pt.Array)
+                return arg
+
+        return pt.einsum(spec, *(preprocess_arg(arg) for arg in args))
+
+    def permits_inplace_modification(self):
```
inducer, line 531:
```suggestion
    @property
    def permits_inplace_modification(self):
```
x-ref: #38, my bad

arraycontext/fake_numpy.py
matthiasdiener: done in https://github.com/inducer/arraycontext/pull/14/commits/609a503f47d2f4bc5e887087992561a5d236ce47

test/test_arraycontext.py
matthiasdiener: Done(?) in https://github.com/inducer/arraycontext/pull/14/commits/95fd7be8645e6d35d71aa903f70d441facc9f0d8

arraycontext/pytest.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/14/commits/14bf6045407ba614f88255d860a5da419c7b8018

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+class _PytatoFakeNumpyLinalgNamespace(BaseFakeNumpyLinalgNamespace):
+    # Everything is implemented in the base class for now.
+    pass
+
+
+class _PytatoFakeNumpyNamespace(BaseFakeNumpyNamespace):
+    """
+    A :mod:`numpy` mimic for :class:`PytatoArrayContext`.
+
+
```
inducer, line 51:
```suggestion
```

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+    .. note::
+
+        :mod:`pytato` does not define any memory layout. If the caller invokes
+        any routine that depends on the input arrays' memory layout, a result is
+        returned assuming the arrays have a C-contiguous layout.
```
inducer, line 56: I disagree with this. Anything that relies on memory layout should be an error, consistent with pytato: https://documen.tician.de/pytato/design.html#memory-layout.

arraycontext/__init__.py
```diff
@@ -47,12 +47,16 @@
         rec_multimap_array_container,
         mapped_over_array_containers,
         multimapped_over_array_containers,
+        keyed_map_array_container,
+        rec_keyed_map_array_container,
```
inducer, line 51: Kind of don't like these enough to have them exported to the main namespace.

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+                    "sqrt", "exp"]
+        if name in pt_funcs:
+            import pytato as pt  # type: ignore
+            from functools import partial
+            return partial(rec_map_array_container, getattr(pt, name))
+
+        return super().__getattr__(name)
+
+    def reshape(self, a, newshape):
+        import pytato as pt
```
inducer, line 74: Move the fake-numpy bitss to a submodule so you don't need to have keep importing this. (Analogous issue for the pyopencl array context: #39)

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+
+
+@dataclass(frozen=True, eq=True)
+class ArrayContainerInputDescriptor(AbstractInputDescriptor):
+    id_to_ary_descr: "PMap[Tuple[Union[str, int], ...], Tuple[np.dtype, \
+                                                        Tuple[int, ...]]]"
+
+
+@dataclass
+class PytatoCompiledOperator:
```
inducer, line 194: Docstring?

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+        self.program_cache[arg_descrs] = PytatoExecutable(self.actx,
+                                                          pytato_program,
+                                                          input_naming_map,
+                                                          output_naming_map,
+                                                          output_template=outputs)
+
+        return self.program_cache[arg_descrs](*args)
+
+
+class PytatoExecutable:
```
inducer, line 290: * Make this a dataclass?
* Add a docstring

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+        # }}}
+
+        def to_output_template(keys, _):
+            return self.actx.thaw(out_dict[self.output_id_to_name_in_program[keys]])
+
+        return rec_keyed_map_array_container(to_output_template,
+                                             self.output_template)
+
+
+class PytatoArrayContext(ArrayContext):
```
inducer, line 362:
```suggestion
class PytatoPyOpenCLArrayContext(ArrayContext):
```
(There's reason to believe that there could be others.)

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+
+        def _rec_ravel(a):
+            if order in "FC":
+                return pt.reshape(a, (-1,), order=order)
+            elif order in "AK":
+                # memory layout is assumed to be "C"
+                return pt.reshape(a, (-1,), order="C")
+            else:
+                raise ValueError("`order` can be one of 'F', 'C', 'A' or 'K'. "
+                                 f"(got {order})")
```
inducer, line 167:
```suggestion
            else:
                raise ValueError("`order` can be one of 'F' or 'C'. "
                                 f"(got {order})")
```

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+
+    .. attribute:: queue
+
+        A :class:`pyopencl.CommandQueue`.
+
+    .. attribute:: allocator
+
+        A :mod:`pyopencl` memory allocator. Can also be None (default) or False
+        to use the default allocator.
+    """
```
inducer, line 375:
```suggestion

    .. automethod:: __init__
    """
```

arraycontext/impl/pyopencl.py
```diff
@@ -338,6 +268,9 @@ def from_numpy(self, array: np.ndarray):
         return cla.to_device(self.queue, array, allocator=self.allocator)
 
     def to_numpy(self, array):
+        from numpy import isscalar
+        if isscalar(array):
```
inducer, line 272: Why this instead of `isinstance(array, numbers.Number)`? (Change or add a comment.)

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+
```
inducer, line 376: We should discuss our strategy for handling of scalars. It seems that the current strategy here is for `PyOpenCLArrayContext` to keep returning plain scalars, where as the pytato array context will return lazy scalars. This is kind of inconsistent, but it might be the best strategy available to us. In the medium term, I would like it if `PyOpenCLArrayContext` started returning device scalars as well, but that's a large-scale compatibility break. #40 for the pyopencl end of this. @kaushikcfd @matthiasdiener basically what I'm looking for here from you is whether you agree with this direction.

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+        import pytato as pt
+        import pyopencl.array as cla
+
+        if isinstance(array, pt.Placeholder):
+            raise ValueError("freezing placeholder would return garbage valued"
+                    " arrays")
+        if isinstance(array, cla.Array):
+            return array.with_queue(None)
+        if not isinstance(array, pt.Array):
+            raise TypeError("PytatoArrayContext.freeze invoked with non-pt "
```
inducer, line 429: > pt

Spell out.

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+        import pyopencl.array as cla
+        from arraycontext import (is_array_container,
+                                  rec_keyed_map_array_container)
+
+        input_kwargs_to_loopy = {}
+
+        # {{{ extract loopy arguments execute the program
+
+        for pos, arg in enumerate(args):
+            if isinstance(arg, np.number):
```
inducer, line 310: Why this vs `isinstance(Number)` vs. `np.isscalar`?

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+                                                 .program.default_entrypoint
+                                                 .arg_dict)}
+
+        # }}}
+
+        evt, out_dict = self.pytato_program(queue=self.actx.queue,
+                                            allocator=self.actx.allocator,
+                                            **input_kwargs_to_loopy)
+
+        evt.wait()
```
inducer, line 351: Why the `wait`?

arraycontext/impl/pytato.py
```diff
@@ -0,0 +1,536 @@
...
+
+
+@dataclass(frozen=True, eq=True)
+class ScalarInputDescriptor(AbstractInputDescriptor):
+    dtype: np.dtype
+
+
+@dataclass(frozen=True, eq=True)
+class ArrayContainerInputDescriptor(AbstractInputDescriptor):
+    id_to_ary_descr: "PMap[Tuple[Union[str, int], ...], Tuple[np.dtype, \
```
inducer, line 189: Type variable, as above.

inducer (commented):
Thanks for working on this. Some comments below.

inducer:

arraycontext/impl/pyopencl.py
matthiasdiener: No particular reason, changed.

arraycontext/impl/pyopencl.py
```diff
@@ -371,10 +325,9 @@ def transform_loopy_program(self, t_unit):
             if "i1" in all_inames:
                 inner_iname = "i1"
         else:
-            raise RuntimeError(
-                "Unable to reason what outer_iname and inner_iname "
-                f"needs to be; all_inames is given as: {all_inames}"
-            )
+            # cannot "fit" the optimization strategy for the provided kernel
+            # => bail
+            return t_unit
```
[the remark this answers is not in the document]
kaushikcfd: Agreed. This was there because of the there wasn't a transformation strategy for the fused-batch kernels. But, given that that's been fixed recently, reverted back to raising.

arraycontext/__init__.py
kaushikcfd: I thought they were helpful routine, but I guess the vagueness of key-spec is a show-stopper: 0980f4b.

arraycontext/impl/pytato.py
kaushikcfd: @inducer: The problem is other array-contexts rely on memory layouts, forcing pytato to have a memory layout. Downstream users also make use of the memory layout information: https://github.com/inducer/meshmode/blob/dab6ecffa61521ba3ab5fc72de2ad2a5fb431960/meshmode/dof_array.py#L670-L671. IMO, it cannot be avoided.

inducer: I guess what I'm objecting to is the globality and specificity of this statement, which I find very restricting. AFAICT, this only applies to `ravel` right now, correct? My preferred solution would be to remove this statement and replace it with a pointer to the one in pytato, and to document `ravel(order="A")` as returning the array elements in "unspecified" order.

kaushikcfd: Thanks for the clarification. I agree, we can just restrict this "unspecified layout" to ravel, rather than globally.

arraycontext/impl/pytato.py
kaushikcfd: Thanks: 48b4679.

arraycontext/impl/pytato.py
kaushikcfd: I agree, I think we should allow both types of scalar i.e. host and device for all operations. But, we should allow a flag which doesn't return host-scalar unless to_numpy-ed.

arraycontext/container/traversal.py
```diff
@@ -233,6 +233,76 @@ def wrapper(*args: Any) -> Any:
...
+            return f(keys, _ary)
+        elif is_array_container(_ary):
+            def _tuple_if_not_tuple(x: Any) -> Tuple[Union[str, int], ...]:
+                if not isinstance(x, tuple):
+                    assert isinstance(x, (str, int))
+                    return x,
+                else:
+                    assert all(isinstance(el, (str, int))
+                               for el in x)
+                    return x
```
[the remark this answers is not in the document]
kaushikcfd: Agreed: 51f5d3d.

arraycontext/impl/pytato.py
matthiasdiener: Done

arraycontext/impl/pytato.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/14/commits/27c6688788637f7d5380db29c55346aa27cb632a

arraycontext/impl/pytato.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/14/commits/e334008d19775abf2fe981284aa6436447e0a409

arraycontext/impl/pytato.py
inducer: I think allowing both types just complicates things. `to_numpy` isn't too hard to type.

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,58 @@ class BaseFakeNumpyLinalgNamespace:
...
+            import numpy.linalg as la
+            return la.norm(
+                    [self.norm(_flatten_cl_array(subary), ord=ord)
+                        for _, subary in serialize_container(ary)],
+                    ord=ord)
+
+        if is_array_container(ary):
+            if ord == np.inf:
+                return max([self.norm(subary, ord=ord)
+                        for _, subary in serialize_container(ary)])
```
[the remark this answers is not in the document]
kaushikcfd: fixed:  66faa71

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,58 @@ class BaseFakeNumpyLinalgNamespace:
...
+        from arraycontext.impl import _is_meshmode_dofarray, _is_pyopencl_array
+
+        if _is_pyopencl_array(ary):
+            if ary.ndim != 1:
+                from arraycontext.impl import _flatten_cl_array
+                # mimics numpy's norm computation
+                return self.norm(_flatten_cl_array(ary), ord=2)
+
+        if (_is_meshmode_dofarray(ary) and all([_is_pyopencl_array(subary)
+                                for _, subary in serialize_container(ary)])):
```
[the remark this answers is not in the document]
kaushikcfd: fixed: 66faa71

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,58 @@ class BaseFakeNumpyLinalgNamespace:
...
+        from numbers import Number
+        if isinstance(ary, Number):
+            return abs(ary)
+
+        if ord is None:
+            ord = 2
+
+        from arraycontext.impl import _is_meshmode_dofarray, _is_pyopencl_array
+
+        if _is_pyopencl_array(ary):
```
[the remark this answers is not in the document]
kaushikcfd: Removed:  66faa71

arraycontext/fake_numpy.py
```diff
@@ -174,6 +174,58 @@ class BaseFakeNumpyLinalgNamespace:
...
+            if ary.ndim != 1:
+                from arraycontext.impl import _flatten_cl_array
+                # mimics numpy's norm computation
+                return self.norm(_flatten_cl_array(ary), ord=2)
```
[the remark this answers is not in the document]
kaushikcfd: Thanks; 66faa71.

arraycontext/container/traversal.py
```diff
@@ -233,6 +233,76 @@ def wrapper(*args: Any) -> Any:
...
+
+            return deserialize_container(_ary, [
+                    (key, frec(keys+_tuple_if_not_tuple(key), subary))
```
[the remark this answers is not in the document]
kaushikcfd: I was earlier shooting for a flattened version of the tuple. But I see that could lead to ambiguous keys. Fixed:  07d3d4b.

arraycontext/container/traversal.py
```diff
@@ -233,6 +233,76 @@ def wrapper(*args: Any) -> Any:
     update_wrapper(wrapper, f)
     return wrapper
 
+
+def keyed_map_array_container(
+        f: Callable[[Union[str, int], Any], Any],
```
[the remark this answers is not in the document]
kaushikcfd: 07d3d4b

arraycontext/container/traversal.py
```diff
@@ -233,6 +233,76 @@ def wrapper(*args: Any) -> Any:
...
+                    ])
+        else:
+            return f(keys, _ary)
+
+    frec = rec if recursive else f
+    return rec((), ary)
+
+
+def rec_keyed_map_array_container(
+        f: Callable[[Tuple[Union[str, int], ...], Any], Any],
```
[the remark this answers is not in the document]
kaushikcfd: Yep, went with `Any`: 07d3d4b

arraycontext/impl/pytato.py
kaushikcfd: 0dc10bd

arraycontext/impl/__init__.py
```diff
@@ -21,3 +25,40 @@
...
+        # Work around https://github.com/inducer/pyopencl/pull/402
+        return ary._new_with_changes(
+                data=None, offset=0, shape=(0,), strides=(ary.dtype.itemsize,))
+    if ary.flags.f_contiguous:
+        return ary.reshape(-1, order="F")
+    elif ary.flags.c_contiguous:
+        return ary.reshape(-1, order="C")
+    else:
+        raise ValueError("cannot flatten array "
+                f"with strides {ary.strides} of {ary.dtype}")
```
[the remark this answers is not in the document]
kaushikcfd: Agreed, thanks: b893cd2.

arraycontext/impl/pytato.py
kaushikcfd: Oops, thanks: c305104.

arraycontext/impl/pytato.py
kaushikcfd: Thanks: c305104.

arraycontext/impl/pytato.py
kaushikcfd: Followed `serialize_array_container` and kept type of keys to be `Any`.

arraycontext/impl/pytato.py
kaushikcfd: For the DAGs I had generally dealt with, they generally had high global temporary variable allocation for each kernel. And so it made sense to flush the queue before enqueuing another kernel. However, looking at `PyOpenCLArrayContext`, I think I should hack-in `wait_event_queue_length`.

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,212 @@
...
+        from pytato.loopy import call_loopy
+        entrypoint, = set(program.callables_table)
+
+        # thaw frozen arrays
+        kwargs = {kw: (self.thaw(arg) if isinstance(arg, cla.Array) else arg)
+                  for kw, arg in kwargs.items()}
+
+        return call_loopy(program, kwargs, entrypoint)
+
+    def freeze(self, array):
```
kaushikcfd, line 98: If `array: pt.DataWrapper`, should we short-circuit the logic here?

arraycontext/container/traversal.py
```diff
@@ -101,14 +101,14 @@ def _multimap_array_container_impl(
...
-                ), f"expected type '{type(template_ary).__name__}'"
-
         if (type(template_ary) is leaf_cls
                 or not is_array_container(template_ary)):
             return f(*_args)
 
+        assert all(
+                type(_args[i]) is type(template_ary) for i in container_indices[1:]
+                ), f"expected type '{type(template_ary).__name__}'"
+
```
kaushikcfd, line 111: To reviewers: this change had to be done because in pytato leaf arrayclasses are not identical. i.e. one entry could be a `pt.IndexLambda`, other could be `pt.Placeholder``

arraycontext/loopy.py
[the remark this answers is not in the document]
kaushikcfd: Hopefully https://github.com/inducer/loopy/pull/431 should help us revert this change.

arraycontext/pytest.py
```diff
@@ -31,9 +31,44 @@
...
+        return ("<Pytato array context factory for <pyopencl.Device '%s' on '%s'>"
+                % (self.device.name.strip(),
+                 self.device.platform.name.strip()))
+
+
+types_to_factories = {PyOpenCLArrayContext: _PyOpenCLArrayContextFactory,
+    PytatoArrayContext: _PytatoArrayContextFactory}
+
+
+def pytest_generate_tests_for_array_contexts(metafunc, actx_list=None) -> None:
```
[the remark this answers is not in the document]
matthiasdiener: Sorry, this still isn't clear to me. Could you please provide more details about what should be returned, and how this would be used? @kaushikcfd came up with the following code that works with the current code, but this isn't what you are looking for I guess?:

```python
pytest_generate_tests = 
  lambda metafunc: pytest_generate_tests_for_array_contexts(metafunc, actx_list=[PytatoPyOpenCLArrayContext])
```

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+                    id_to_ary_descr[keys] = (np.dtype(ary.dtype),
+                                             ary.shape)
+                    return ary
+
+                rec_keyed_map_array_container(id_collector, arg)
+                return ArrayContainerInputDescriptor(pmap(id_to_ary_descr))
+            else:
+                raise ValueError("Argument to a compiled operator should be"
+                                 " either a scalar or an array container. Got"
+                                 f" '{arg}'.")
```
inducer, line 123: Could this function be moved outside? (Maybe with an underscore?)

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+                input_naming_map[(pos, )] = name
+                return pt.make_placeholder((), np.dtype(arg), name)
+            elif is_array_container(arg):
+                def _rec_to_placeholder(keys, ary):
+                    name = (f"_actx_in_{pos}_"
+                            + _ary_container_key_stringifier(keys))
+                    input_naming_map[(pos,) + keys] = name
+                    return pt.make_placeholder(ary.shape, ary.dtype,
+                                               name)
+                return rec_keyed_map_array_container(_rec_to_placeholder,
```
inducer, line 154: This calls `rec_keyed_map_array_container` twice, i.e. it uses two (expensive) traversals of the input tree even in the case of a cache hit. At some expense to code clarity, I think the same effect could be achieved in one traversal. IMO that's worth it, because this is a fairly "hot" code path.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+            else:
+                raise NotImplementedError(type(arg))
+
+        outputs = self.f(*[to_placeholder(arg, iarg)
+                           for iarg, arg in enumerate(args)])
+
+        if not is_array_container(outputs):
+            # TODO: We could possibly just short-circuit this interface if the
+            # returned type is a scalar. Not sure if it's worth it though.
+            raise ValueError(f"Function '{self.f.__name__}' to be compiled did not"
```
inducer, line 165:
```suggestion
            raise NotImplementedError(f"Function '{self.f.__name__}' to be compiled did not"
```
(I think we'll want this before too long.)

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+        elif isinstance(key, tuple):
+            return "tup" + "_".join(_rec_str(k) for k in key) + "endtup"
+        else:
+            raise NotImplementedError
+
+    return "_".join(_rec_str(key) for key in keys)
+
+
+@dataclass
+class PytatoCompiledOperator:
```
inducer, line 78:
```suggestion
class LazilyCompilingFunctionCaller:
```

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+                                                          pytato_program,
+                                                          input_naming_map,
+                                                          output_naming_map,
+                                                          output_template=outputs)
+
+        return self.program_cache[arg_descrs](*args)
+
+
+@dataclass
+class PytatoExecutable:
```
inducer, line 192:
```suggestion
class CompiledFunction:
```

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+                return ArrayContainerInputDescriptor(pmap(id_to_ary_descr))
+            else:
+                raise ValueError("Argument to a compiled operator should be"
+                                 " either a scalar or an array container. Got"
+                                 f" '{arg}'.")
+
+        arg_descrs = tuple(to_arg_descr(arg) for arg in args)
+
+        try:
+            exec_f = self.program_cache[arg_descrs]
```
inducer, line 128:
```suggestion
            compiled_f = self.program_cache[arg_descrs]
```

inducer (commented):
Thanks for continuing to work on this! A few more thoughts below. (I had... dodged... the difficult bit in my previous review.)

arraycontext/impl/pytato/compile.py
kaushikcfd: Yep, thanks: f618fa4.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class AbstractInputDescriptor:
+    def __eq__(self, other):
+        raise NotImplementedError
+
+    def __hash__(self):
+        raise NotImplementedError
+
+
```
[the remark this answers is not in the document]
kaushikcfd: Yep thanks: 97df833

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+import numpy as np
+from typing import Any, Callable, Tuple, Dict
+from dataclasses import dataclass, field
+from pyrsistent import pmap, PMap
+
+import pyopencl.array as cla
+import pytato as pt
+
+
+class AbstractInputDescriptor:
```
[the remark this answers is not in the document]
kaushikcfd: 97df833

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+
+@dataclass(frozen=True, eq=True)
+class ScalarInputDescriptor(AbstractInputDescriptor):
+    dtype: np.dtype
+
+
+@dataclass(frozen=True, eq=True)
+class ArrayContainerInputDescriptor(AbstractInputDescriptor):
+    id_to_ary_descr: "PMap[Tuple[Any, ...], Tuple[np.dtype, \
+                                                  Tuple[int, ...]]]"
```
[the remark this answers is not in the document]
kaushikcfd: Agreed, that tuple might "grow": 97df833

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def _ary_container_key_stringifier(keys: Tuple[Any, ...]) -> str:
+    """
+    Helper for :meth:`PytatoCompiledOperator.__call__`. Stringifies an
+    array-container's component's key. The aim is that no-two different keys
+    have the same stringification.
+    """
```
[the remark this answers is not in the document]
kaushikcfd: See a68ec30.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+    def _rec_str(key: Any) -> str:
+        if isinstance(key, (str, int)):
+            return str(key)
+        elif isinstance(key, tuple):
+            return "tup" + "_".join(_rec_str(k) for k in key) + "endtup"
+        else:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
kaushikcfd: See a68ec30.

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,212 @@
...
+        import pytato as pt
+        import pyopencl.array as cla
+
+        if isinstance(array, pt.Placeholder):
+            raise ValueError("freezing placeholder would return garbage valued"
+                    " arrays")
```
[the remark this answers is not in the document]
kaushikcfd: > Please tell me that this is not actually what would happen if this conditional were removed. (Instead, I would expect the loopy program to complain of missing arguments.)

Yep that would happen.

> As error checking goes, this is incomplete: If a Placeholder instance is problematic, 2*Placeholder(...) is just as problematic.

Oops yep.

Removed this check, as I agree the loopy invoker would handle this. See 8545978

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,212 @@
...
+        prg = lp.set_options(prg, "insert_additional_gbarriers")
+
+        return gridify(prg)
+
+    def tag(self, tags: Union[Sequence[Tag], Tag], array):
+        # Sorry, not capable.
+        return array
+
+    def tag_axis(self, iaxis, tags: Union[Sequence[Tag], Tag], array):
+        # Sorry, not capable.
```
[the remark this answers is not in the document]
kaushikcfd: Done in e07ceb6

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,212 @@
...
+            # }}}
+
+            return knl
+
+        prg = lp.set_options(prg, "insert_additional_gbarriers")
+
+        return gridify(prg)
+
+    def tag(self, tags: Union[Sequence[Tag], Tag], array):
+        # Sorry, not capable.
```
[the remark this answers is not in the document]
kaushikcfd: Thanks, yep: e07ceb6

arraycontext/impl/pytato/compile.py
kaushikcfd: Agreed. Fixed in 5145a56.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+class ArrayContainerInputDescriptor(AbstractInputDescriptor):
+    id_to_ary_descr: "PMap[Tuple[Any, ...], Tuple[np.dtype, \
+                                                  Tuple[int, ...]]]"
+
+
+def _ary_container_key_stringifier(keys: Tuple[Any, ...]) -> str:
+    """
+    Helper for :meth:`PytatoCompiledOperator.__call__`. Stringifies an
+    array-container's component's key. The aim is that no-two different keys
+    have the same stringification.
```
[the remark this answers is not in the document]
kaushikcfd: Yep, sure! 7005678.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,283 @@
...
+    f: Callable[..., Any]
+    program_cache: Dict[Tuple[AbstractInputDescriptor, ...],
+                        "PytatoExecutable"] = field(default_factory=lambda: {})
+
+    def __call__(self, *args: Any) -> Any:
+        """
+        Mimics :attr:`~PytatoCompiledOperator.f` being called with *args*.
+        Before calling :attr:`~PytatoCompiledOperator.f`, it is compiled to a
+        :mod:`pytato` DAG that would apply :attr:`~PytatoCompiledOperator.f`
+        with *args* in a lazy-sense.
```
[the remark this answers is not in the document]
kaushikcfd: How does b2bd3df look?

arraycontext/pytest.py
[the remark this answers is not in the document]
matthiasdiener: Is #44 a potential solution for this?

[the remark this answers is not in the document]
kaushikcfd: Yep, I think so.

matthiasdiener:
This is ready for review I think.

kaushikcfd:
> This is ready for review I think.

I still need to take care of one comment. I will update here,  once that's done.

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,212 @@
...
+                    knl = lp.split_iname(knl, f"{bigger_loop}_inner",
+                            nwi[0], inner_tag="l.1")
+                    knl = lp.split_iname(knl, smaller_loop,
+                            nwi[1], inner_tag="l.0")
+                elif isinstance(insn, lp.BarrierInstruction):
+                    pass
+                else:
+                    raise NotImplementedError
+
+            # }}}
```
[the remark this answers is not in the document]
inducer: #50 deprecated this on the pyopencl front, this logic really belongs in meshmode. (And it should use the transform metadata introduced in https://github.com/inducer/meshmode/pull/236 instead of matching names.)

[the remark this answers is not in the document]
matthiasdiener: Done in 430c8d8

arraycontext/container/traversal.py
kaushikcfd: Sure, done 433826e.

kaushikcfd:
This is ready to go in from end too!

matthiasdiener:
Ready for review @inducer.

The followup PR with meshmode's version of transform_loopy_program is here: https://github.com/inducer/meshmode/pull/216

arraycontext/impl/pyopencl.py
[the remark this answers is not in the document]
kaushikcfd: > fused-batch kernels.

To elaborate on this: on the meshmode end, the interpolation batch kernels had to be fused so that those operations could be supported by array-contexts that do not allow in-place modification of arrays. We had this "do-nothing" transformation strategy for those kernels as custom-iname tagging didn't become a thing until recently.

Given that that's been fixed on the meshmode end, we reverted it back to raising an error when no-transform strategy could be found.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+
+        rec_keyed_map_array_container(_as_dict_of_named_arrays,
+                                      outputs)
+
+        pytato_program = pt.generate_loopy(dict_of_named_arrays,
+                                           options={"return_dict": True},
+                                           cl_device=self.actx.queue.device)
+
+        pytato_program.program = self.actx.transform_loopy_program(pytato_program
+                                                                   .program)
```
inducer, line 225: Stateful: Copy with new program instead?

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # {{{ extract loopy arguments execute the program
+
+        for arg_id, arg in arg_id_to_arg.items():
+            if np.isscalar(arg):
+                arg = cla.to_device(self.actx.queue, np.array(arg))
+            elif isinstance(arg, pt.array.DataWrapper):
+                arg = arg.data
+            elif isinstance(arg, cla.Array):
+                pass
```
inducer, line 288: What's going on here? (Add a comment to explain maybe?)

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+                                                output_template=outputs)
+
+        return self.program_cache[arg_id_to_descr](arg_id_to_arg)
+
+
+@dataclass
+class CompiledFunction:
+    """
+    A callable which is an instance of :attr:`~LazilyCompilingFunctionCaller.f`
+    specialized for a particular input type fed to it.
```
inducer, line 239:
```suggestion
    A callable which captures the :class:`pytato.target.BoundProgram`  resulting
    from calling :attr:`~LazilyCompilingFunctionCaller.f` with a given set of
    input types, and generating :mod:`loopy` IR from it.```

**Review by inducer — COMMENTED** [review]
Thanks! This is on the home stretch. Mostly some simple stuff to clean up, then this is ready to go.

File: arraycontext/impl/pyopencl.py (hunk 37 above)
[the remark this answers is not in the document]
Comment (inducer) [reply arraycontext/impl/pyopencl.py]: OIC, thanks for explaining.

File: arraycontext/impl/pytato/fake_numpy.py [hunk 70]
```diff
@@ -0,0 +1,149 @@
...
+
+
+class PytatoFakeNumpyNamespace(BaseFakeNumpyNamespace):
+    """
+    A :mod:`numpy` mimic for :class:`PytatoPyOpenCLArrayContext`.
+
+    .. note::
+
+        :mod:`pytato` does not define any memory layout for its arrays. See
+        `Pytato docs <https://documen.tician.de/pytato/design.html#memory-layout>`_
```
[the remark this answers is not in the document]
kaushikcfd: b8de351

arraycontext/pytest.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, device):
         """
         self.device = device
 
+    @property
+    def actx_class(self):
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
kaushikcfd: e515c2d

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+        to use the default allocator.
+
+    .. automethod:: __init__
+    """
+
+    def __init__(self, queue, allocator=None):
+        super().__init__()
+        self.queue = queue
+        self.allocator = allocator
+        self.np = self._get_fake_numpy_namespace()
```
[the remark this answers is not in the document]
kaushikcfd: ef0914c

arraycontext/impl/pytato/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # unused, but necessary to keep the context alive
+        self.context = self.queue.context
+
+    def _get_fake_numpy_namespace(self):
+        from arraycontext.impl.pytato.fake_numpy import PytatoFakeNumpyNamespace
+        return PytatoFakeNumpyNamespace(self)
+
```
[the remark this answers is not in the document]
kaushikcfd: ef0914c

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+                                           ) -> "Tuple[PMap[Tuple[Any, ...],\
+                                                            Any],\
+                                                       PMap[Tuple[Any, ...],\
+                                                            AbstractInputDescriptor]\
+                                                       ]":
+    """
+    Helper for :meth:`LazilyCompilingFunctionCaller.__call__`. Extracts the
+    argument id to argument values and descriptor mappings from the input
+    arguments. See :attr:`CompiledFunction.input_id_to_name_in_program` for
+    argument-id's representation.
```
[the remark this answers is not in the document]
kaushikcfd: a80b924

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+@dataclass
+class LazilyCompilingFunctionCaller:
+    """
+    Records a side-effect-free callable :attr:`LazilyCompilingFunctionCaller.f`, that
+    would be specialized for different input types
+    :meth:`LazilyCompilingFunctionCaller.__call__` is invoked with.
+
+    .. attribute:: f
+
+        The callable that would be specialized into :mod:`pytato` DAGs.
```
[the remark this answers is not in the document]
kaushikcfd: 7ec975b

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+
+    return pmap(arg_id_to_arg), pmap(arg_id_to_descr)
+
+
+@dataclass
+class LazilyCompilingFunctionCaller:
+    """
+    Records a side-effect-free callable :attr:`LazilyCompilingFunctionCaller.f`, that
+    would be specialized for different input types
+    :meth:`LazilyCompilingFunctionCaller.__call__` is invoked with.
```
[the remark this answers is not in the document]
kaushikcfd: 7ec975b

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+            else:
+                raise NotImplementedError(type(arg))
+
+            input_kwargs_to_loopy[self.input_id_to_name_in_program[arg_id]] = arg
+
+        evt, out_dict = self.pytato_program(queue=self.actx.queue,
+                                            allocator=self.actx.allocator,
+                                            **input_kwargs_to_loopy)
+
+        evt.wait()
```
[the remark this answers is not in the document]
kaushikcfd: 5935dea

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+                return ary
+
+            rec_keyed_map_array_container(id_collector, arg)
+        else:
+            raise ValueError("Argument to a compiled operator should be"
+                             " either a scalar or an array container. Got"
+                             f" '{arg}'.")
+
+    for iarg, arg in enumerate(args):
+        to_arg_descr(iarg, arg)
```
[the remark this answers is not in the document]
kaushikcfd: Now that I look back, I agree. Seem unnecessary: fcda4c8.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+    arguments. See :attr:`CompiledFunction.input_id_to_name_in_program` for
+    argument-id's representation.
+    """
+    arg_id_to_arg: Dict[Tuple[Any, ...], Any] = {}
+    arg_id_to_descr: Dict[Tuple[Any, ...], AbstractInputDescriptor] = {}
+
+    def to_arg_descr(iarg: int, arg: Any) -> None:
+        """
+        Returns a :class:`AbstractInputDescriptor` for a
+        attr:`LazilyCompilingFunctionCaller.f`'s input argument.
```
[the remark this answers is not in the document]
kaushikcfd: fcda4c8

arraycontext/impl/pytato/__init__.py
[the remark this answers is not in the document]
kaushikcfd: Had to revert in 5224c16 as `super()` is the one that sets `self.np` using `_get_fake_numpy_namespace`.

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+
+        outputs = self.f(*[to_placeholder(arg, iarg)
+                           for iarg, arg in enumerate(args)])
+
+        if not is_array_container(outputs):
+            # TODO: We could possibly just short-circuit this interface if the
+            # returned type is a scalar. Not sure if it's worth it though.
+            raise NotImplementedError(
+                f"Function '{self.f.__name__}' to be compiled "
+                f"did not return an array container, but '{outputs}' instead.")
```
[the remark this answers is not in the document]
kaushikcfd: d3c8279

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+
+        dict_of_named_arrays = {}
+        # output_naming_map: result id to name of the named array in the
+        # generated pytato DAG.
+        output_naming_map = {}
+        # input_naming_map: argument id to placeholder name in the generated
+        # pytato DAG.
+        input_naming_map = {}
+
+        def to_placeholder(arg, pos):
```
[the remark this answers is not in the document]
kaushikcfd: Rewired in 696e653. to avoid closures. Hopefully that improved its maintainability.

arraycontext/impl/pytato/compile.py
kaushikcfd: Will do after https://github.com/inducer/pytato/pull/114.

arraycontext/impl/pytato/compile.py
kaushikcfd: Thanks, 4080e44

arraycontext/impl/pytato/compile.py
```diff
@@ -0,0 +1,308 @@
...
+            if np.isscalar(arg):
+                name = f"_actx_in_{pos}"
+                input_naming_map[(pos, )] = name
+                return pt.make_placeholder((), np.dtype(arg), name)
+            elif is_array_container(arg):
+                def _rec_to_placeholder(keys, ary):
```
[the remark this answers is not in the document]
kaushikcfd: I'm guessing this was a concern about the closure being unreadable rather than a perf. hit. That's fixed in 696e653.

doc/array_context.rst
```diff
@@ -12,3 +12,9 @@ Array context based on :mod:`pyopencl.array`
 --------------------------------------------
 
 .. automodule:: arraycontext.impl.pyopencl
+
+
+Array context based on :mod:`pytato`
```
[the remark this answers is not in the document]
kaushikcfd: How do you feel about: 5ddb2e6?

test/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+from arraycontext import (  # noqa: F401
+        pytest_generate_tests_for_array_contexts,
+        )
+
+import logging
+logger = logging.getLogger(__name__)
+
+pytest_generate_tests = pytest_generate_tests_for_array_contexts([
+    "pyopencl", "pyopencl-deprecated", "pytato-pyopencl"
+    ])
```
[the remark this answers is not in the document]
kaushikcfd: Yep, was unnecessary. 616aa12.

arraycontext/pytest.py
```diff
@@ -234,14 +260,17 @@ def inner(metafunc):
...
         metafunc.parametrize(arg_names, arg_value_tuples, ids=ids)
 
     return inner
 
 
 def pytest_generate_tests_for_pyopencl_array_context(metafunc) -> None:
     """Parametrize tests for pytest to use a
-    :class:`~arraycontext.PyOpenCLArrayContext`.
+    :class:`~arraycontext.ArrayContext`.
```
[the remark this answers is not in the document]
kaushikcfd: Sorry, fixed in e6602ba. I'm guessing it was probably a merge leftover that got overlooked.

kaushikcfd:
For anyone wondering about the CI failures: https://github.com/inducer/meshmode/pull/238 should fix those.

arraycontext/impl/pytato/compile.py
kaushikcfd: Done, see  d23600f.

doc/array_context.rst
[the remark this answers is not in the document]
inducer: :+1:

arraycontext/impl/pytato/compile.py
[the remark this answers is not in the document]
inducer: It was. Sorry about the snarkiness. I like the new version much better, thanks for accommodating me.

arraycontext/impl/pytato/compile.py
inducer: Yup, does. Thanks!

setup.py
```diff
@@ -44,7 +44,6 @@ def main():
             "loopy>=2019.1",
             "dataclasses; python_version<'3.7'",
             "types-dataclasses",
-            "pytato>=2021.1",
```
inducer: Thanks! I agree... no need for a hard dependency. `arraycontext` works just fine without pytato.

inducer:
Thanks all for seeing this through. Looks great! In it goes.

PR #15, opened 2021-05-28
Implement, test container-over-container broadcasting

### LINKED ISSUE #12: Container-over-container broadcasting, broadcasting rules
As pointed out by @mtcam in https://github.com/illinois-ceesd/mirgecom/pull/355#discussion_r637478702:
```
E TypeError: unsupported operand type(s) for *: 'DOFArray' and 'ConservedVars'
```
(Here, `DOFArray`s clearly want to be inside `ConservedVars`, so there must be a rule to specify who is the "outer" structure.)

The biggest question here is which type should broadcast over which other type, and how the rules are to be communicated.

Some proposals:
* Just give the outer container's `with_container_arithmetic` a `bcast_container_types=(type1, type2)` argument.
* Define a `_container_broadcast_priority`: the higher the "outermost". All containers broadcast over all other containers (or at least ones that have this enabled in `with_container_arithmetic`. Kind of like numpy's `__array_priority__`.
* Have containers have a method `may_broadcast_over_container_type(other_type)`.

Some thoughts:
* This is defining a partial order of who broadcasts over who. How do we ensure that this stays consistent? (An easy way to get an inconsistent situation is to tell two types that they should each be the "outer" type with respect to each other. That's easy to catch, but scenarios with three or more "edges" in the cycle are also possible and harder to catch.) Also, maybe it's not necessary to catch this statically...? Joys of multiple dispatch. What does Julia do?
* Technically, object arrays are part of this hierarchy. `bcast_obj_array` [here](https://documen.tician.de/arraycontext/container.html#arraycontext.with_container_arithmetic), unfortunately, casually defines them as "outer". I think we'll have to deprecate that argument, barely a week after it was introduced...
* Of the three proposals above, I *think* I like `bcast_container_types` the best. The priority thing feels restrictive and ill-defined. `may_broadcast_over_container_type` is probably slow.
* Maybe we'll need ways to specify both "inner" and "outer" types. `DOFArray` has no reason to know about `ConservedVars` (so it shouldn't have to know who its outer types are). Meanwhile, if we let object arrays particpate in the hierarchy, then many types will want them as their outermost type, but we can't ask object arrays who their inner types are.

Thoughts?

cc @alexfikl @thomasgibson

inducer:
Closes #12

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -37,6 +37,7 @@ import numpy as np
 
 class _OpClass(enum.Enum):
     ARITHMETIC = enum.auto
+    MATMUL = enum.auto
     BITWISE = enum.auto
     SHIFT = enum.auto
     EQ_COMPARISON = enum.auto
@@ -58,6 +59,8 @@ _BINARY_OP_AND_DUNDER = [
         ("pow", "{} ** {}", True, _OpClass.ARITHMETIC),
         ("mod", "{} % {}", True, _OpClass.ARITHMETIC),
         ("divmod", "divmod({}, {})", True, _OpClass.ARITHMETIC),
+
+        ("matmul", "{} @ {}", True, _OpClass.MATMUL),
 
         ("and", "{} & {}", True, _OpClass.BITWISE),
         ("or", "{} | {}", True, _OpClass.BITWISE),
@@ -112,10 +115,10 @@ def _format_binary_op_str(op_str, arg1, arg2):
         return op_str.format(arg1, arg2)
 
 
-def with_container_arithmetic(
+def with_container_arithmetic(*,
         bcast_number=True, bcast_obj_array=None, bcast_numpy_array=False,
         bcast_container_types=None,
-        arithmetic=True, bitwise=False, shift=False,
+        arithmetic=True, matmul=False, bitwise=False, shift=False,
         eq_comparison=None, rel_comparison=None):
     """A class decorator that implements built-in operators for array containers
     by propagating the operations to the elements of the container.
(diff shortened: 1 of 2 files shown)


arraycontext/container/arithmetic.py
```diff
@@ -122,6 +126,14 @@ def with_container_arithmetic(
...
     :arg bcast_numpy_array: If *True*, any :class:`numpy.ndarray` will broadcast
         over the container.  (with the container as the 'inner' structure)
+        If this is set to *True*, *bcast_obj_array* must also be *True*.
+    :arg bcast_container_types: A sequence of container types that will broadcast
+        over this container (with this container as the 'outer' structure).
+        :class:`numpy.ndarray` is permitted to be part of this sequence to
+        indicate that, in such broadcasting situations, this container should
+        be the 'outer' structure.)
+        In this case, *bcast_obj_array* (and consequently *bcast_numpy_array*)
+        must be *False*.
```
alexfikl, line 136:
```suggestion
        be the 'outer' structure. In this case, *bcast_obj_array* 
        (and consequently *bcast_numpy_array*) must be *False*.
```

alexfikl (approved):
Looks good to me!

How about that case with `flux @ normal` broadcasting? Would that just require adding `__matmul__` to `_BINARY_OP_AND_DUNDER`? Although that should really just work on `ndarray`..

inducer:
> How about that case with `flux @ normal` broadcasting? Would that just require adding `__matmul__` to `_BINARY_OP_AND_DUNDER`? 

Good catch! Implemented and tested in 5622da4. [The notation](https://github.com/inducer/arraycontext/blob/5622da4da9576c371122bc545be776d461a8e5a7/test/test_arraycontext.py#L587-L596) comes out looking bearable IMO.

> Although that should really just work on `ndarray`..

I can't think of any other leaf types that implement `__matmul__`. Nonetheless, conceivably the operator-type arguments to `with_container_arithmetic` could conceivably allow type tuples to make them each more picky. I'll save that for when we need it though.

PR #18, opened 2021-06-07
Enable mypy

alexfikl:


inducer (approved):
Thanks! LGTM

inducer:
(Grudge downstream might fail, because we're in the middle of the https://github.com/inducer/loopy/pull/222 merge.)

PR #22, opened 2021-06-09
Improve freeze/thaw usability

### LINKED ISSUE #19: Check that array expressions don't mix array contexts
(at least optionally)

See also #21.

cc @thomasgibson

### LINKED ISSUE #21: Better explain the freeze/thaw concept
x-ref: https://github.com/inducer/grudge/pull/121#discussion_r648598039

cc @thomasgibson

inducer:
In addition to a better explanation of `freeze` and `thaw`, this PR makes it possible to fail noisily in situations like https://github.com/inducer/grudge/pull/121#discussion_r648598039. Specifically, the new `ArrayContext.clone()` method allows making the "setup" array context of the discretization semantically distinct from the array context used in user code, which allows these mistakes to be caught.

This PR comes with additional small changes in meshmode and grudge that make use of these features:
* https://github.com/inducer/meshmode/pull/218
* https://github.com/inducer/grudge/pull/123

@thomasgibson Could you take a look, specifically at the explanation that I've added?

Closes #19.
Closes #21.

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -133,7 +133,7 @@ def with_container_arithmetic(
         matmul: bool = False,
         bitwise: bool = False,
         shift: bool = False,
-        _same_cls_check: Optional[str] = None,
+        _cls_has_array_context_attr: bool = False,
         eq_comparison: Optional[bool] = None,
         rel_comparison: Optional[bool] = None) -> Callable[[type], type]:
     """A class decorator that implements built-in operators for array containers
@@ -161,10 +161,10 @@ def with_container_arithmetic(
     :arg rel_comparison: If *True*, implement ``<``, ``<=``, ``>``, ``>=``.
         In that case, if *eq_comparison* is unspecified, it is also set to
         *True*.
-    :arg _same_cls_check: A string of Python code operating on ``arg1`` and ``arg2``
-        to perform additional checks in the case where both arguments are of the
-        type of the decorated class.
-        Only inserted if :data:`__debug__` is *True*.
+    :arg _cls_has_array_context_attr: A flag indicating whether the decorated
+        class has an ``array_context`` attribute. If so, and if :data:`__debug__`
+        is *True*, an additional check is performed in binary operators
+        to ensure that both containers use the same array context.
         Consider this argument an unstable interface. It may disappear at any moment.
 
     Each operator class also includes the "reverse" operators if applicable.
@@ -327,8 +327,11 @@ def with_container_arithmetic(
             with Indentation(gen):
                 gen("if arg2.__class__ is cls:")
                 with Indentation(gen):
-                    if __debug__ and _same_cls_check:
-                        gen(_same_cls_check)
+                    if __debug__ and _cls_has_array_context_attr:
+                        gen("""
+                            if arg1.array_context is not arg2.array_context:
+                                raise ValueError("array contexts of both arguments "
+                                    "must match")""")
                     gen(f"return cls({zip_init_args})")
                 gen(f"""
                 if {bool(outer_bcast_type_names)}:  # optimized away
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -133,6 +133,7 @@ def with_container_arithmetic(
         matmul: bool = False,
         bitwise: bool = False,
         shift: bool = False,
+        _same_cls_check: Optional[str] = None,
         eq_comparison: Optional[bool] = None,
         rel_comparison: Optional[bool] = None) -> Callable[[type], type]:
     """A class decorator that implements built-in operators for array containers
@@ -160,6 +161,11 @@ def with_container_arithmetic(
     :arg rel_comparison: If *True*, implement ``<``, ``<=``, ``>``, ``>=``.
         In that case, if *eq_comparison* is unspecified, it is also set to
         *True*.
+    :arg _same_cls_check: A string of Python code operating on ``arg1`` and ``arg2``
+        to perform additional checks in the case where both arguments are of the
+        type of the decorated class.
+        Only inserted if :data:`__debug__` is *True*.
+        Consider this argument an unstable interface. It may disappear at any moment.
 
     Each operator class also includes the "reverse" operators if applicable.
 
@@ -245,7 +251,7 @@ def with_container_arithmetic(
                     "'_deserialize_init_arrays_code'. If this is a dataclass, "
                     "use the 'dataclass_array_container' decorator first.")
 
-        from pytools.codegen import CodeGenerator
+        from pytools.codegen import CodeGenerator, Indentation
         gen = CodeGenerator()
         gen("""
             from numbers import Number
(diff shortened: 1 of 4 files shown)


inducer:
@anderson2981 @mtcam I recall that both of you also had gripes about the freeze/thaw interface. Please take a look here and see if this helps make this a bit clearer.

inducer:
I'm starting to like this more and more. It found a few bugs in meshmode, for instance:
- https://github.com/inducer/meshmode/pull/218/checks?check_run_id=2788506274
- https://github.com/inducer/meshmode/pull/218/commits/77342161958a6801e83fa5dcdc30e73fef7bd54a

arraycontext/container/arithmetic.py
```diff
@@ -133,6 +133,7 @@ def with_container_arithmetic(
         matmul: bool = False,
         bitwise: bool = False,
         shift: bool = False,
+        _same_cls_check: Optional[str] = None,
```
alexfikl, line 136: Not liking the idea of letting users inject arbitrary code in here (even more than with `*_init_arrays_code`, I guess?). Why not just always enforce the array context check?

arraycontext/context.py
```diff
@@ -2,6 +2,69 @@
...
+
+*   "Thawed" arrays are associated with an array context. They use that context
+    to carry out operations (arithmetic, function calls).
+
+*   "Frozen" arrays are static data. They are not associated with an array context,
+    and no operations can be performed on them.
+
+Freezing and thawing may be used to move arrays from one array context to another,
+as long as both array contexts use identical in-memory data representation.
+
```
alexfikl, line 23:
```suggestion
as long as both array contexts use identical in-memory data representation.
Otherwise, a common format must be agreed upon, for example using 
:mod:`numpy` through :meth:`~arraycontext.ArrayContext.to_numpy` and
:meth:`~arraycontext.ArrayContext.from_numpy`.
```
?

arraycontext/context.py
```diff
@@ -2,6 +2,69 @@
...
+    of long-lived data that should be frozen.
+
+-   Within a function, if the user did not supply an array context,
+    then any data returned to the user should be frozen. Note that, array contexts
+    need not necessarily be a separate argument.
+
+-   Simply supplying thawed data suffices to provide an array context, using
+    e.g.  :func:`~arraycontext.get_container_context` or
+    :func:`~arraycontext.get_container_context_recursively`; the array context
+    need to be passed in as a separate function argument.
```
alexfikl, line 37:
```suggestion
    need not be passed in as a separate function argument.
```
?

arraycontext/context.py
```diff
@@ -2,6 +2,69 @@
...
+    such a command queue is necessary; it is the main means of synchronization
+    structure between the host program and the compute device. "Thawing" here
+    means associating an array with a command queue, and "freezing" means
+    ensuring that the array data is fully computed in memory and
+    decoupling the array from the command queue. It is not valid to "mix"
+    arrays associated with multiple queues within an operation: if it were allowed,
+    a dependent operation might begin computing before an input is fully
+    available. (Since bugs of this nature would be very difficult to
+    find, :class:`pyopencl.array.Array` and
+    :class:`~meshmode.dof_array.DOFArray` will not allow them.
```
alexfikl, line 57:
```suggestion
    :class:`~meshmode.dof_array.DOFArray` will not allow them.)
```

arraycontext/context.py
```diff
@@ -2,6 +2,69 @@
...
+
+-   For the lazily-evaluating array context based on :mod:`pytato`,
+    "thawing" corresponds to the creation of a placeholder representing
```
alexfikl, line 60: Link to some placeholder docs?

arraycontext/context.py
```diff
@@ -259,6 +322,25 @@ def einsum(self, spec, *args, arg_names=None, tagged=()):
...
+    def clone(self):
+        """If possible, return a version of *self* that is semantically
+        equivalent (i.e. implements all array operations in the same way)
+        but is a separate object. May return *self* if that is not possible.
+
+        .. note::
+
+            The main objective of this semi-documented method is to help
+            flag errors more clearly when array contexts are mixed that
+            shouldn't be. For example, at the time of this writing,
```
alexfikl, line 335:
```suggestion
            should not be. For example, at the time of this writing,
```

arraycontext/context.py
```diff
@@ -259,6 +322,25 @@ def einsum(self, spec, *args, arg_names=None, tagged=()):
...
+            :class:`meshmode.meshmode.Discretization` objects have a private
+            array context that is only to be used only for setup-related tasks.
```
alexfikl, line 337:
```suggestion
            array context that is only to be used for setup-related tasks.
```

alexfikl (approved):
Looks good to me!

* I'm liking the `clone` thing to ensure that array contexts don't mix when they shouldn't!
* Can't say much about the added docs, but left a few nitpicks here and there.

arraycontext/container/arithmetic.py
inducer: Yeah, it's yucky and ad-hoc, hence the underscoring. Maybe you can tell I wasn't super proud of the interface. :)

At the same time, it's not generically possible to find the array context from the array. (That's the price we pay for allowing un-wrapped array types like the pyopencl array or the numpy array.) Calling `get_container_context` would run into the same problem. I also would like to keep this code path fast. Those are the constraints from my point of view. I would absolutely welcome better ideas, but I'm not sure I have any.

alexfikl: How about adding a `promise_this_class_has_an_array_context_attribute: bool = False` and generating the check based on that? (preferably with a better name)

arraycontext/context.py
inducer: 912fda1

inducer: (I.e. was that about what you had in mind?)

inducer:
@alexfikl Thanks for taking a look!

arraycontext/context.py
alexfikl: Yeah, that looks good!

arraycontext/container/arithmetic.py
inducer: That is better. 21b1462

alexfikl: Nice! :tada:

arraycontext/context.py
```diff
@@ -1,4 +1,71 @@
 """
+Freezing and thawing
```
MTCam, line 2: 👍 This is great.  One thing that might be cool is a corresponding blurb in *MIRGE-Com* with a link to this documentation, including some examples of how to use it in *MIRGE-Com* and when it is neeeded.

arraycontext/context.py
```diff
@@ -1,4 +1,71 @@
...
+    of long-lived data that should be frozen.
+
+-   Within a function, if the user did not supply an array context,
+    then any data returned to the user should be frozen. Note that, array contexts
+    need not necessarily be a separate argument.
+
+-   Simply supplying thawed data suffices to provide an array context, using
+    e.g.  :func:`~arraycontext.get_container_context` or
+    :func:`~arraycontext.get_container_context_recursively`; the array context
+    need not be passed in as a separate function argument.
```
thomasgibson, line 36: I like this. A lot. In fact, could be section this out via:
```suggestion
Usage guidelines
^^^^^^^^^^^^^^^^
Here are some rules of thumb to use when dealing with thawing and freezing:

-   Any array that is stored for a long time needs to be frozen.
    "Memoized" data (cf. :func:`pytools.memoize` and friends) is a good example
    of long-lived data that should be frozen.

-   Within a function, if the user did not supply an array context,
    then any data returned to the user should be frozen. Note that, array contexts
    need not necessarily be a separate argument.

-   Simply supplying thawed data suffices to provide an array context, using
    e.g.  :func:`~arraycontext.get_container_context` or
    :func:`~arraycontext.get_container_context_recursively`; the array context
    need not be passed in as a separate function argument.
```

thomasgibson: Just to make it stick out more

thomasgibson (approved):
I think the docs are extremely helpful. My only comment really is to emphasize the guidelines. I think this PR is basically merge-ready as is though

arraycontext/context.py
inducer: Good idea. I added some anchors we can use to link in 91fdaa6.

arraycontext/context.py
```diff
@@ -1,4 +1,76 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Freezing and thawing are abstract names for concrete operations. It may be helpful
+to understand what these operations mean in the concrete case of various
+actual array contexts:
+
+-   Each :class:`~arraycontext.PyOpenCLArrayContext` is associated with a
+    :class:`pyopencl.CommandQueue`. In order to operate on array data,
+    such a command queue is necessary; it is the main means of synchronization
+    structure between the host program and the compute device. "Thawing" here
```
inducer, line 53:
```suggestion
    between the host program and the compute device. "Thawing" here
```

PR #29, opened 2021-06-13
Kernel transform strategy based on iname tags

inducer:
This is intended to help revive https://github.com/inducer/meshmode/pull/217. It's used in https://github.com/inducer/meshmode/pull/220.

I'm not sure this has any business being here, but I didn't have the stamina to give meshmode its own `PyOpenCLArrayContext` subclass again. Nonetheless, I think that's what we should do. @matthiasdiener, I'd be grateful for your help here.

 That means:
- updating those imports across the tests and examples (in meshmode), and
- making the pytest fixture customizable as to what actx classes to use,
- Same goes for grudge and mirgecom, as they will likely grow their own kernel metadata at some point. 
- This also makes the other transform strategies redundant, so those can really go. (and the other meshmode kernels should be updated to use this one)
- So `PyOpenCLArrayContext.transform_loopy` here should really just raise. To still be able to run tests for this package, introduce a subclass that has a no-op `transform_loopy`.
-  To avoid a huge performance pitfall, it's important that the array context provides a useful error if no applicable transform strategy was found.

Draft because:
- [x] Needs https://github.com/inducer/loopy/pull/410
- [x] Point `requirements.txt` back to `main`
- [ ] Should really live in meshmode.

cc @thomasgibson @alexfikl

### DIFF
--- a/arraycontext/impl/pyopencl.py
+++ b/arraycontext/impl/pyopencl.py
@@ -347,11 +347,29 @@ class PyOpenCLArrayContext(ArrayContext):
         all_inames = default_entrypoint.all_inames()
         # FIXME: This could be much smarter.
         inner_iname = None
-        if (len(default_entrypoint.instructions) == 1
-                and isinstance(default_entrypoint.instructions[0], lp.Assignment)
+
+        from arraycontext.metadata import ElementInameTag, DOFInameTag
+        el_inames = [iname
+                for iname in default_entrypoint.inames.values()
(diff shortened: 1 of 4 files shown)


inducer:
If you tried this and dismissed it: cfa98e7 fixed a big problem with the transform cache. (not sure how I missed that)

inducer:
Superseded by https://github.com/inducer/meshmode/pull/236.

PR #30, opened 2021-06-14
Mimic numpy's norm when ord is None

kaushikcfd:
- When ord is None, numpy ravels the array and then computes the euclidean norm.

I couldn't find a mention for this in the numpy docs, but here is a small script:
```python
In [2]: a = np.random.rand(3, 3, 3, 3, 3)

In [3]: np.linalg.norm(a)
Out[3]: 8.718438810223871

In [4]: np.linalg.norm(a.ravel(), ord=2)
Out[4]: 8.718438810223871
```

Let me know what do you guys think.

### DIFF
--- a/arraycontext/impl/pyopencl.py
+++ b/arraycontext/impl/pyopencl.py
@@ -155,9 +155,9 @@ def _flatten_array(ary):
         return ary._new_with_changes(
                 data=None, offset=0, shape=(0,), strides=(ary.dtype.itemsize,))
     if ary.flags.f_contiguous:
-        return ary.reshape(-1, order="F")
+        return ary.ravel(order="F")
     elif ary.flags.c_contiguous:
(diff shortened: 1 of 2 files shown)


inducer:
Thanks! If you add a test showing that the behavior matches numpy, I could see going along with this. @alexfikl?

kaushikcfd:
> If you add a test showing that the behavior matches numpy

Done! See 1bae084.

alexfikl:
> Thanks! If you add a test showing that the behavior matches numpy, I could see going along with this. @alexfikl?

Yeah, this should match the [docs for linalg.norm](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html)
```
If both axis and ord are None, the 2-norm of x.ravel will be returned.
```

inducer:
LGTM. Thanks!

PR #34, opened 2021-06-14
Implement PyOpenCLArrayContext.ravel

kaushikcfd:


arraycontext/impl/pyopencl.py
```diff
@@ -141,6 +142,25 @@ def concatenate(self, arrays, axis=0):
             self._array_context.allocator
         )
 
+    def ravel(self, a, order="C"):
+        def _rec_ravel(a):
+            if order == "K":
```
alexfikl, line 147: What about `order == "A"`? Raise a `NotImplementedError`?

https://numpy.org/doc/stable/reference/generated/numpy.ravel.html#numpy.ravel

alexfikl: Actually, from the docs it seems like `"A"` is the one that does Fortran order or C order depending on which one is contiguous. Am I misreading that?

kaushikcfd: Yeah this should really be "A" and not implemented for "K".

kaushikcfd: Thanks, see 0e11990.

PR #35, opened 2021-06-14
Special-case serialization of 1/2D object arrays

inducer:
@alexfikl Too gross? On a test case based on meshmode's `simple-dg`, this goes from ~11.4s to ~11.2s.

### DIFF
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -172,7 +172,16 @@ def _serialize_ndarray_container(ary: np.ndarray) -> Iterable[Tuple[Any, Any]]:
         raise ValueError(
                 f"only object arrays are supported, given dtype '{ary.dtype}'")
 
-    return np.ndenumerate(ary)
+    # special-cased for speed
+    if ary.ndim == 1:
+        return [(i, ary[i]) for i in range(ary.shape[0])]
+    elif ary.ndim == 2:
+        return [((i, j), ary[i, j])
+                for i in range(ary.shape[0])
+                for j in range(ary.shape[1])
+                ]
+    else:
+        return np.ndenumerate(ary)
 
 
 @deserialize_container.register(np.ndarray)


alexfikl (approved):
It's not intrusive and simple enough to not bother me much at all, so if the speedup is worth it on your end :+1:!

arraycontext/container/__init__.py
```diff
@@ -172,7 +172,16 @@ def _serialize_ndarray_container(ary: np.ndarray) -> Iterable[Tuple[Any, Any]]:
         raise ValueError(
                 f"only object arrays are supported, given dtype '{ary.dtype}'")
 
-    return np.ndenumerate(ary)
+    # special-cased for speed
+    if ary.ndim == 1:
+        return [(i, ary[i]) for i in range(ary.shape[0])]
```
kaushikcfd, line 177: Shouldn't the key be a tuple
```python
In [4]: list(np.ndenumerate(np.random.rand(3)))
Out[4]: 
[((0,), 0.09292195880848719),
 ((1,), 0.6780962855552635),
 ((2,), 0.3681587065690297)]
```

kaushikcfd: I.e.:
```python
In [6]: ary = np.random.rand(3)

In [7]: list(np.ndenumerate(ary))
Out[7]: 
[((0,), 0.646399478266698),
 ((1,), 0.015079989468580912),
 ((2,), 0.3586538930864147)]

In [8]: [(i, ary[i]) for i in range(ary.shape[0])]
Out[8]: [(0, 0.646399478266698), (1, 0.015079989468580912), (2, 0.3586538930864147)]
```

Was this intentional or a typo?

inducer: Sorry... it merged before I had a chance to see your comment. IMO it's OK. The types of the keys are defined by the individual object, the array container widgetry doesn't care.

And for the object array deserialization, all that matters is that the key works as an index for an array of the same shape, which it does:

https://github.com/inducer/arraycontext/blob/ef662bec5f829e2586c328127d259a3347c54a86/arraycontext/container/__init__.py#L196-L197

PR #37, opened 2021-06-17
Add a flag to indicate whether an array context permits in-place modification

inducer:
The context for this is that some operations, specifically the `DirectConnection` kernels of https://github.com/inducer/meshmode/pull/220, benefit quite a bit from being able to do in-place updates, see https://github.com/inducer/meshmode/pull/220#issuecomment-861927006. So the idea would be to include both code paths and use in-place updates only if allowed. For testing, we'd of course make a way to force both code paths to be used. Adding this feels a bit ad-hoc, OTOH I think it's quite possibly a reasonable thing for an array context to tell a user.

cc @thomasgibson

### DIFF
--- a/arraycontext/context.py
+++ b/arraycontext/context.py
@@ -103,7 +103,7 @@ THE SOFTWARE.
 """
 
 from typing import Sequence, Union
-from abc import ABC, abstractmethod
+from abc import ABC, abstractmethod, abstractproperty
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


inducer:
Spoke with @alexfikl, I think we agree that it's not beautiful, but it may be worth it for the moment. Let's see where it goes without fully committing to it.

PR #41, opened 2021-06-18
Import pyopencl.array once at the top

### LINKED ISSUE #39: Do not import PyOpenCLArrayContext into the root namespace
Move `PyOpenCLFakeNumpyNamespace` (and the linalg bit) to a submodule so that we can import `pyopencl` globally there and avoid all these repetitive, performance-killing imports
https://github.com/inducer/arraycontext/blob/322c18fe1983c6a03a7c86bf0ec3fb401422353a/arraycontext/impl/pyopencl.py#L114-L127

while still avoiding a hard dependency on pyopencl.

kaushikcfd:
Closes #39.

### DIFF
--- a/arraycontext/impl/pyopencl.py
+++ b/arraycontext/impl/pyopencl.py
@@ -43,6 +43,12 @@ from arraycontext.container.traversal import (rec_multimap_array_container,
 from arraycontext.container import serialize_container, is_array_container
 from arraycontext.context import ArrayContext
 
+try:
+    import pyopencl as cl  # noqa: F401
+    import pyopencl.array as cl_array
+except ImportError:
+    pass
+
 
 # {{{ fake numpy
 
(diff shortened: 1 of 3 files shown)


arraycontext/impl/pyopencl.py
```diff
@@ -43,6 +43,12 @@
 from arraycontext.container import serialize_container, is_array_container
 from arraycontext.context import ArrayContext
 
+try:
+    import pyopencl as cl  # noqa: F401
+    import pyopencl.array as cl_array
+except ImportError:
+    pass
```
inducer, line 50: Thanks for thinking about this. I don't think this suffices, unfortunately. Importing pyopencl is relatively heavyweight, and we should wait to do it until the user asks us to do so. As a result, I think the best option is to move the numpy bits to a separate file.

kaushikcfd: But wouldn't we end up doing:
```
from arraycontext.impl.pyopencl.fake_numpy import maximum, ...
```
everywhere? i.e. I think we will still keep on adding a line before every math routine invocation.

inducer: Not sure I understand. In my view, the only place that imports from that submodule would be the `.np` property.

kaushikcfd: Aah right. I understand and I think this will work. Thanks!

inducer:
Thanks! LGTM.

PR #43, opened 2021-06-23
Force device scalars on PyOpenCLArrayContext

### LINKED ISSUE #40: Scalars in PyOpenCLArrayContext should be device scalars
This is for consistency with the pytato context and for performance: if all that's needed is a device scalar, there's no real need for the device/host/device round-trip.

Unfortunately, that's a rather larger compatibility break. One way to do this would be to introduce a flag on the CL array context that changes the behavior, and whose old default value is deprecated. The pytato array context kind of leads the charge here anyway: it requires that `to_numpy` is called on lazy scalars before they're "normally" usable.

cc @kaushikcfd @matthiasdiener

alexfikl:
Added a flag with a deprecated default value as suggested in #40.

- [x] Test `force_device_scalars=True` too.

Fixes #40.

### DIFF
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -156,8 +156,7 @@ class PyOpenCLArrayContext(ArrayContext):
         return cl_array.to_device(self.queue, array, allocator=self.allocator)
 
     def to_numpy(self, array):
-        from numbers import Number
-        if not self._force_device_scalars and isinstance(array, Number):
+        if not self._force_device_scalars and np.isscalar(array):
             return array
 
         return array.get(queue=self.queue)
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -28,7 +28,7 @@ THE SOFTWARE.
 """
 
 from warnings import warn
-from typing import Any, Dict, List, Sequence, Optional, Union, TYPE_CHECKING
+from typing import Dict, List, Sequence, Optional, Union, TYPE_CHECKING
 
 import numpy as np
 
@@ -117,7 +117,7 @@ class PyOpenCLArrayContext(ArrayContext):
 
         self._force_device_scalars = force_device_scalars
         self._wait_event_queue_length = wait_event_queue_length
-        self._kernel_name_to_wait_event_queue: Dict[str, List[Any]] = {}
+        self._kernel_name_to_wait_event_queue: Dict[str, List[cl.Event]] = {}
 
         if queue.device.type & cl.device_type.GPU:
             if allocator is None:
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -133,7 +133,7 @@ class PyOpenCLArrayContext(ArrayContext):
                         "a noticeable speed improvement.")
 
         self._loopy_transform_cache: \
-                Dict[lp.TranslationUnit, lp.TranslationUnit] = {}
+                Dict["lp.TranslationUnit", "lp.TranslationUnit"] = {}
 
     def _get_fake_numpy_namespace(self):
         from arraycontext.impl.pyopencl.fake_numpy import PyOpenCLFakeNumpyNamespace
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -2,6 +2,7 @@
 .. currentmodule:: arraycontext
 .. autoclass:: PyOpenCLArrayContext
 """
+
 __copyright__ = """
 Copyright (C) 2020-1 University of Illinois Board of Trustees
 """
@@ -27,7 +28,7 @@ THE SOFTWARE.
 """
 
 from warnings import warn
-from typing import Sequence, Union
+from typing import Any, Dict, List, Sequence, Optional, Union, TYPE_CHECKING
 
 import numpy as np
 
@@ -35,6 +36,10 @@ from pytools.tag import Tag
 
 from arraycontext.metadata import FirstAxisIsElementsTag
 from arraycontext.context import ArrayContext
+
+
+if TYPE_CHECKING:
+    import pyopencl
 
 
 # {{{ PyOpenCLArrayContext
@@ -62,7 +67,11 @@ class PyOpenCLArrayContext(ArrayContext):
         as the allocator can help avoid this cost.
     """
 
-    def __init__(self, queue, allocator=None, wait_event_queue_length=None):
+    def __init__(self,
+            queue: "pyopencl.CommandQueue",
+            allocator: Optional["pyopencl.tools.AllocatorInterface"] = None,
+            wait_event_queue_length: Optional[int] = None,
+            force_device_scalars: bool = False) -> None:
         r"""
         :arg wait_event_queue_length: The length of a queue of
             :class:`~pyopencl.Event` objects that are maintained by the
@@ -84,18 +93,24 @@ class PyOpenCLArrayContext(ArrayContext):
             For now, *wait_event_queue_length* should be regarded as an
             experimental feature that may change or disappear at any minute.
         """
+        if not force_device_scalars:
+            warn("Returning host scalars from the array context is deprecated. "
+                    "To return device scalars set 'force_device_scalars=True'. "
+                    "Support for returning host scalars will be removed in 2022.",
+                    DeprecationWarning, stacklevel=2)
+
         import pyopencl as cl
 
         super().__init__()
         self.context = queue.context
         self.queue = queue
         self.allocator = allocator if allocator else None
-
         if wait_event_queue_length is None:
             wait_event_queue_length = 10
 
+        self._force_device_scalars = force_device_scalars
         self._wait_event_queue_length = wait_event_queue_length
-        self._kernel_name_to_wait_event_queue = {}
+        self._kernel_name_to_wait_event_queue: Dict[str, List[Any]] = {}
 
         if queue.device.type & cl.device_type.GPU:
             if allocator is None:
@@ -110,7 +125,7 @@ class PyOpenCLArrayContext(ArrayContext):
                         "are running Python in debug mode. Use 'python -O' for "
                         "a noticeable speed improvement.")
 
-        self._loopy_transform_cache = {}
+        self._loopy_transform_cache: dict = {}
 
     def _get_fake_numpy_namespace(self):
         from arraycontext.impl.pyopencl.fake_numpy import PyOpenCLFakeNumpyNamespace
@@ -133,6 +148,10 @@ class PyOpenCLArrayContext(ArrayContext):
         return cl_array.to_device(self.queue, array, allocator=self.allocator)
 
     def to_numpy(self, array):
+        from numbers import Number
+        if not self._force_device_scalars and isinstance(array, Number):
+            return array
+
         return array.get(queue=self.queue)
 
     def call_loopy(self, t_unit, **kwargs):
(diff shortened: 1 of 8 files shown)


alexfikl:
Not quite sure how to add the `force_device_scalars=True` case to the tests as well. Any suggestions?

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -62,7 +67,11 @@ class PyOpenCLArrayContext(ArrayContext):
         as the allocator can help avoid this cost.
     """
 
-    def __init__(self, queue, allocator=None, wait_event_queue_length=None):
+    def __init__(self,
+            queue: "pyopencl.CommandQueue",
+            allocator: Optional["pyopencl.tools.AllocatorInterface"] = None,
+            wait_event_queue_length: Optional[int] = None,
+            force_device_scalars: bool = False) -> None:
```
kaushikcfd, line 74: Docs for `force_device_scalars`?

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -110,7 +125,7 @@ def __init__(self, queue, allocator=None, wait_event_queue_length=None):
                         "are running Python in debug mode. Use 'python -O' for "
                         "a noticeable speed improvement.")
 
-        self._loopy_transform_cache = {}
+        self._loopy_transform_cache: dict = {}
```
kaushikcfd, line 128:
```suggestion
        self._loopy_transform_cache: Dict[lp.TranslationUnit, lp.TranslationUnit]  = {}
```

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -133,6 +148,10 @@ def from_numpy(self, array: np.ndarray):
         return cl_array.to_device(self.queue, array, allocator=self.allocator)
 
     def to_numpy(self, array):
+        from numbers import Number
+        if not self._force_device_scalars and isinstance(array, Number):
```
kaushikcfd, line 152: I think this doesn't work for a `boolean` scalar. Probably `np.isscalar(array)`?

arraycontext/pytest.py
```diff
@@ -59,10 +59,12 @@ def pytest_generate_tests_for_pyopencl_array_context(metafunc):
     from pyopencl.tools import _ContextFactory
 
     class ArrayContextFactory(_ContextFactory):
-        def __call__(self):
+        def __call__(self, force_device_scalars=False):
```
kaushikcfd, line 62: How about PyOpenCLArrayContext always yields 2 variants (like it does with the multiple CL-contexts) one with `force_device_scalars=True` and vice-versa.

kaushikcfd (commented):
Thanks, looks great! Added some minor comments.

arraycontext/impl/pyopencl/__init__.py
alexfikl: Added some docs in 4758865.

arraycontext/impl/pyopencl/__init__.py
alexfikl: Added in 16254e4.

arraycontext/impl/pyopencl/__init__.py
alexfikl: Good point! Added in b20ac9a.

arraycontext/pytest.py
alexfikl: Not a big fan of that at the moment because it's going to force everyone using `pytest_generate_tests_for_pyopencl_array_context` to handle both cases (and break most of the downstream CIs). It should be something to opt-in for.

Maybe something closer to what you did in #14 to allow multiple array contexts?

alexfikl: Gave that a go in #44. I can squash that into this or just merge it after, since this is probably more urgent for the `to_numpy` scalar fix.

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -83,19 +93,31 @@ def __init__(self, queue, allocator=None, wait_event_queue_length=None):
...
         self.queue = queue
         self.allocator = allocator if allocator else None
-
         if wait_event_queue_length is None:
             wait_event_queue_length = 10
 
+        self._force_device_scalars = force_device_scalars
         self._wait_event_queue_length = wait_event_queue_length
-        self._kernel_name_to_wait_event_queue = {}
+        self._kernel_name_to_wait_event_queue: Dict[str, List[Any]] = {}
```
inducer, line 120:
```suggestion
        self._kernel_name_to_wait_event_queue: Dict[str, List[pyopencl.Event]] = {}
```
and cast below, to keep the types informative?

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -37,6 +38,11 @@
 from arraycontext.context import ArrayContext
 
 
+if TYPE_CHECKING:
+    import pyopencl
```
inducer, line 42:
```suggestion
    import pyopencl as cl
```
?

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -133,6 +156,9 @@ def from_numpy(self, array: np.ndarray):
         return cl_array.to_device(self.queue, array, allocator=self.allocator)
 
     def to_numpy(self, array):
+        if not self._force_device_scalars and np.isscalar(array):
```
inducer, line 159: Why `isscalar` over `isinstance(..., Number)`?

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -175,6 +186,21 @@ def _flatten_array(ary):
...
 
+def _cl_scalar_list_norm(ary, ord):
+    if ord is None:
+        ord = 2
+
+    from numbers import Number
+    if ord == np.inf:
+        return max(ary)
+    elif ord == -np.inf:
+        return min(ary)
```
inducer, line 197: Huh, TIL. (OK to resolve when seen.)

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -175,6 +186,21 @@ def _flatten_array(ary):
...
+    elif isinstance(ord, Number) and ord > 0:
+        return sum(iary**ord for iary in ary)**(1/ord)
+    else:
+        raise NotImplementedError(f"unsupported value of 'ord': {ord}")
+
+
```
inducer, line 203: #14 contains a refactor that moves the norm into the base class. Maybe nab that, so that we only have to touch this once?

arraycontext/pytest.py
inducer: Agree with @alexfikl here. If we change the behavior of this, it'll break tons of downstream code without warning. The whole point of the flag is to allow the change to be gradual and opt-in.

Also, thanks for taking on #44, too. Will review that next, but given the below, I suspect I'd prefer them going in via the same PR.

test/test_arraycontext.py
```diff
@@ -180,8 +180,10 @@ def assert_close_to_numpy_in_containers(actx, op, args):
             ("where", 3),
             ("conj", 1),
             ])
-def test_array_context_np_workalike(actx_factory, sym_name, n_args):
-    actx = actx_factory()
+@pytest.mark.parametrize("force_device_scalars", [True, False])
+def test_array_context_np_workalike(actx_factory, sym_name, n_args,
+        force_device_scalars):
+    actx = actx_factory(force_device_scalars)
```
inducer, line 186: This isn't ideal IMO; it changes the actx factory "interface" in an implementation-specific way. I assume that's what #44 is about.

inducer (commented):
Thanks for taking this on! I agree that the sooner we get this done, the better.

arraycontext/impl/pyopencl/__init__.py
alexfikl: Updated in 08b4199.

Not quite sure what you mean about the cast? mypy doesn't seem to complain, so I assumed it's fine.

arraycontext/impl/pyopencl/__init__.py
alexfikl: 08b4199

arraycontext/impl/pyopencl/fake_numpy.py
alexfikl: numpy seems to support this, so why not! I doesn't correspond to any actual normed space (to my knowledge), but it may be useful to bound things.

arraycontext/impl/pyopencl/__init__.py
alexfikl: It also catches `np.bool_` and `np.bool8`.

https://github.com/inducer/arraycontext/pull/43#discussion_r656733216

test/test_arraycontext.py
alexfikl: Yep, it's merged in now!

arraycontext/impl/pyopencl/fake_numpy.py
alexfikl: Moved in 2bc76bb with the changes from #14 (hopefully got everything).

alexfikl:
@inducer This should be ready for another review round.

arraycontext/pytest.py
```diff
@@ -28,10 +31,208 @@
...
 
+from typing import Any, Callable, Dict, Optional, Sequence, Type, Union
+
+import pyopencl as cl
+from arraycontext.context import ArrayContext
+
+
+# {{{ array context factories
+
+class PytestArrayContextFactory:
```
inducer, line 42:
```suggestion
class PytestPyOpenCLArrayContextFactory:
```

inducer (commented):
Thanks for working on this! Just a few things on the testing end, otherwise ready to go.

arraycontext/pytest.py
```diff
@@ -28,10 +31,208 @@
...
+    force_device_scalars = False
+
+
+_DEFAULT_ARRAY_CONTEXT_FACTORY_DICT: Dict[str, Type[PytestArrayContextFactory]] = {
+        "pyopencl": _PyOpenCLArrayContextFactory,
+        }
+_ARRAY_CONTEXT_FACTORY_DICT: Dict[str, Type[PytestArrayContextFactory]] = {
+        "pyopencl-deprecated": _DeprecatedPyOpenCLArrayContextFactory,
+        }
+_ARRAY_CONTEXT_FACTORY_DICT.update(_DEFAULT_ARRAY_CONTEXT_FACTORY_DICT)
```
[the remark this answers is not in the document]
alexfikl: Renamed in d6c5300.

arraycontext/pytest.py
```diff
@@ -28,10 +31,208 @@
...
+# }}}
+
 
 # {{{ pytest integration
 
-def pytest_generate_tests_for_pyopencl_array_context(metafunc):
+def pytest_generate_tests_for_array_contexts(
+        factories: Optional[Sequence[
+            Union[str, Type[PytestArrayContextFactory]]
+            ]] = None,
```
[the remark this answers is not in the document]
alexfikl: Agreed and updated in d6c5300 (also remove the `_DEFAULT` variable).

arraycontext/pytest.py
```diff
@@ -28,10 +31,208 @@
...
+        ) -> Callable[[Any], None]:
```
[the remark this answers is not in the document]
alexfikl: Sure thing, added in d6c5300.

arraycontext/pytest.py
```diff
@@ -28,10 +31,208 @@
...
+            unique_factories = set(
+                    _DEFAULT_ARRAY_CONTEXT_FACTORY_DICT     # type: ignore[arg-type]
+                    .values())
+        else:
+            unique_factories = set(factories)               # type: ignore[arg-type]
+            unknown_factories = [
+                    factory for factory in unique_factories
+                    if (isinstance(factory, str)
+                        and factory not in _ARRAY_CONTEXT_FACTORY_DICT)
+                    ]
```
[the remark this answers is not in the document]
alexfikl: Yep, should be a bit nicer in d6c5300.

arraycontext/pytest.py
alexfikl: Renamed in d6c5300.

inducer:
Looks great! Thanks for getting to this so quickly.

PR #44, opened 2021-06-23
Extend pytest_generate_tests_for

alexfikl:
This is built on top of #43, so it has a few annoying commits in there (see alexfikl/arraycontext#1 for a nicer full diff). Only a305c3c and 32be014 should be in here.

It adds a `pytest_generate_tests_for_array_context` that takes an `impls` argument, similar to what was proposed in #14. It should be easy to add other array contexts as well, so hopefully it can be used for the lazy one too.

### DIFF
--- a/arraycontext/pytest.py
+++ b/arraycontext/pytest.py
@@ -1,8 +1,9 @@
 """
 .. currentmodule:: arraycontext
+
+.. autofunction:: pytest_generate_tests_for_array_context
 .. autofunction:: pytest_generate_tests_for_pyopencl_array_context
 """
-
 
 __copyright__ = """
 Copyright (C) 2020-1 University of Illinois Board of Trustees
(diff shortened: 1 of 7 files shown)


arraycontext/pytest.py
```diff
@@ -28,48 +29,96 @@
...
+_ALL_ARRAY_CONTEXT_FACTORY_DICT.update(_ARRAY_CONTEXT_FACTORY_DICT)
+
+# }}}
+
+
+# {{{ pytest integration
+
+def pytest_generate_tests_for_array_context(
+        metafunc,
+        impls: Optional[Sequence[str]] = None):
```
kaushikcfd, line 76: I like this, however I would prefer the type to be `Optional[Sequence[Union[str, ArrayContext]]]`. A use case for this would be say meshmode decides to go with its own `ArrayContext` implementation for transformations that are specific to it, then they would want to add that in here.

kaushikcfd (approved):
Thanks, I quite like this over the one in #44, as the tests are test scripts are much cleaner now.

arraycontext/pytest.py
alexfikl: That wouldn't work because it needs to be able to instantiate an `ArrayContext` from a `cl.Device` (at least the way it's currently all set up). How about adding an undocumented
```
def register_array_context_factory(
        name: str,
        func: Callable[[Any], Callable[[], ArrayContext]]) -> None:
```
? Too complicated? I quite like the string-only bit because you can select them from the command line.

inducer:
Thanks! Left a review at https://github.com/alexfikl/arraycontext/pull/1#pullrequestreview-691195063.

arraycontext/pytest.py
inducer: Also responding to https://github.com/alexfikl/arraycontext/pull/1#discussion_r657513197:

I'm OK with allowing strings for brevity and CLI access, but I think I would like custom, user provided factories to also be allowed. (`register_...` ultimately is based on global shared state, it could lead to name clashes, ...)

alexfikl: To respond here too, documented a barebones interface for user provided factories in a77105e. This should allow adding pyopencl-friendly array contexts that we need for the moment.

alexfikl:
I'll go ahead and merge this back into #43 due to https://github.com/inducer/arraycontext/pull/43#discussion_r657496291. All the comments in alexfikl/arraycontext#1 should be addressed (or at least ready for another review).

PR #45, opened 2021-06-23
Make broadcast settings visible on users of with_container_arithmetic

inducer:
This is to support use cases like https://github.com/illinois-ceesd/mirgecom/pull/331, where the information that `with_container_arithmetic` has about the desired broadcast behavior of a type would also be useful to have. To make it available, this just shoves it into some (undocumented, for now) class attributes. In the linked use case, this would help (e.g.) avoid the need for hardcoding `DOFArray`. Instead, it could check whether `np.ndarray` is in `type(obj)._outer_bcast_names` and then follow that behavior.

I also have a distinct feeling that this might also be a bad idea. @alexfikl, got a gut feeling on this one?

cc @majosm

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -280,7 +280,7 @@ def with_container_arithmetic(
             else:
                 return "(%s,)" % ", ".join(t)
 
-        gen(f"cls._outer_bcast_names = {tup_str(outer_bcast_type_names)}")
+        gen(f"cls._outer_bcast_types = {tup_str(outer_bcast_type_names)}")
         gen(f"cls._bcast_numpy_array = {bcast_numpy_array}")
         gen(f"cls._bcast_obj_array = {bcast_obj_array}")
         gen("")
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -394,7 +394,6 @@ def with_container_arithmetic(
 
         # }}}
 
-        print(gen.get())
         # This will evaluate the module, which is all we need.
         code = gen.get().rstrip()+"\n"
         result_dict = {"_MODULE_SOURCE_CODE": code, "cls": cls}
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -280,6 +280,10 @@ def with_container_arithmetic(
             else:
                 return "(%s,)" % ", ".join(t)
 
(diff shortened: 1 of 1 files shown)


alexfikl:
Is it worth moving the `outer` function somewhere in `arraycontext.container`? That way it would keep all of this a bit more contained if we ever decide to change it.

Besides that, no particular feeling of impending doom on my side :shrug: 

(There's also a stray `print` left in there)

arraycontext/container/arithmetic.py
```diff
@@ -390,6 +394,7 @@ def {fname}(arg2, arg1):
 
         # }}}
 
+        print(gen.get())
```
alexfikl, line 397:
```suggestion
```

alexfikl (approved):

inducer:
> Is it worth moving the `outer` function somewhere in `arraycontext.container`?

I'd be all for that. @majosm? (Added you to the repo just in case.)

inducer:
Thanks for taking a look!

arraycontext/container/arithmetic.py
```diff
@@ -280,6 +280,10 @@ def tup_str(t: Tuple[str, ...]) -> str:
             else:
                 return "(%s,)" % ", ".join(t)
 
+        gen(f"cls._outer_bcast_names = {tup_str(outer_bcast_type_names)}")
```
inducer, line 283:
```suggestion
        gen(f"cls._outer_bcast_types = {tup_str(outer_bcast_type_names)}")
```

majosm:
> > Is it worth moving the `outer` function somewhere in `arraycontext.container`?
> 
> I'd be all for that, especially since it'd help avoid use of private attributes across package boundaries. @majosm? (Added you to the repo just in case.)

Just wondering: how come the specifications in `with_container_arithmetic` don’t become public? Would it be useful for external code to have access to that so it can build on top of the existing mapping functionality?

inducer:
> Just wondering: how come the specifications in `with_container_arithmetic` don’t become public? Would it be useful for external code to have access to that so it can build on top of the existing mapping functionality?

I could very well see it being useful. I just have commitment anxiety about the interface... it seems kind of janky and poorly defined.

majosm:
@inducer Where should I stick `outer`?

inducer:
`container.traversal`, and importable from the root?

majosm:
#46

inducer:
Closing in favor of #46.

PR #46, opened 2021-06-25
Add alternate outer product

majosm:


test/test_arraycontext.py
```diff
@@ -738,6 +738,57 @@ def test_norm_ord_none(actx_factory, ndim):
...
+    assert outer(a_dof, b_dof) == a_dof*b_dof
+
+    # Scalar and vector
+    assert np.all(outer(a_dof, b_ary_of_dofs) == a_dof*b_ary_of_dofs)
+    assert np.all(outer(a_ary_of_dofs, b_dof) == a_ary_of_dofs*b_dof)
+
+    # Two vectors
+    assert (
+        np.all(outer(a_ary_of_dofs, b_ary_of_dofs)
+        == np.outer(a_ary_of_dofs, b_ary_of_dofs)))
```
inducer, line 762: Do these `==` do what you intend? (I'm worried about something being array-valued and then just evaluating to True on account of that.)

* I think we'll want a function that checks structural equality in the presence of array containers.
* We should make sure that array container behavior is safe when it comes to equality comparison.

arraycontext/container/traversal.py
```diff
@@ -522,4 +526,44 @@ def to_numpy(ary: Any, actx: ArrayContext) -> Any:
...
+    object if either/both of *a* and *b* are scalars (whereas :func:`numpy.outer`
+    always returns a matrix). Here the definition of "scalar" includes
+    all non-array-container types and any scalar-like array container types.
+
+    If *a* and *b* are both array containers, the result will have the same type
+    as *a*. If both are array containers and neither is an object array, they must
+    have the same type.
+    """
+
+    def is_scalar(x: Any) -> bool:
```
inducer, line 546:
```suggestion
    def should_treat_as_scalar(x: Any) -> bool:
```

arraycontext/container/arithmetic.py
```diff
@@ -303,6 +303,10 @@ def tup_str(t: Tuple[str, ...]) -> str:
             else:
                 return "(%s,)" % ", ".join(t)
 
+        gen(f"cls._outer_bcast_types = {tup_str(outer_bcast_type_names)}")
+        gen(f"cls._bcast_numpy_array = {bcast_numpy_array}")
+        gen(f"cls._bcast_obj_array = {bcast_obj_array}")
+        gen("")
```
inducer, line 309:
```suggestion
        gen("")

```

arraycontext/container/traversal.py
```diff
@@ -522,4 +526,44 @@ def to_numpy(ary: Any, actx: ArrayContext) -> Any:
...
+        if is_array_container(x):
+            return (
+                not isinstance(x, np.ndarray)
+                and np.ndarray not in x.__class__._outer_bcast_types)
```
inducer, line 550:
```suggestion
                # This condition is whether "ndarrays should broadcast inside x".
                and np.ndarray not in x.__class__._outer_bcast_types)
```

arraycontext/container/traversal.py
```diff
@@ -522,4 +526,44 @@ def to_numpy(ary: Any, actx: ArrayContext) -> Any:
 
 # }}}
 
+
+# {{{ algebraic operations
+
+def outer(a: Any, b: Any) -> Any:
+    """
+    Compute the outer product of *a* and *b*.
```
inducer, line 534:
```suggestion
    Compute the outer product of *a* and *b* while allowing either of them
    to be an :class:`ArrayContainer`.
```

arraycontext/container/traversal.py
```diff
@@ -522,4 +526,44 @@ def to_numpy(ary: Any, actx: ArrayContext) -> Any:
...
+        if is_array_container(x):
+            return (
+                not isinstance(x, np.ndarray)
+                and np.ndarray not in x.__class__._outer_bcast_types)
+        else:
+            return True
+
+    if is_scalar(a) or is_scalar(b):
+        return a*b
+    elif isinstance(a, np.ndarray) and isinstance(b, np.ndarray):
```
inducer, line 556: Narrow to only object arrays in view of #95

majosm: Fixed. (3c9da3a)

inducer:
LGTM, thanks!

PR #51, opened 2021-06-28
Broadcast array arg in binary ops if it's a valid leaf array type

### LINKED ISSUE #49: Broadcasting rules distinguish between host scalar and device scalars
Consider this script:
```python
actx = _acf()

@with_container_arithmetic(
        bcast_obj_array=True,
        bcast_numpy_array=True,
        rel_comparison=True,
        _cls_has_array_context_attr=True)
@dataclass_array_container
@dataclass(frozen=True)
class Foo:
    u: DOFArray

foo = Foo(DOFArray(actx, (actx.zeros(3, dtype=np.float64) + 41, )))

print(foo)                                 # prints: Foo(u=DOFArray((cl.Array([41., 41., 41.]),)))
print(foo + 1)                             # prints: Foo(u=DOFArray((cl.Array([42., 42., 42.]),)))
print(foo + actx.from_numpy(np.ones(())))  # ERROR! 😲
```

I don't see a good reason why we don't support this.

I think we should also go a step ahead and make sure that we broadcast every cl.Array to be added (if legal) with all the leaf arrays of the array container. One way I could see this working is if every `ArrayContext` comes with a `ARRAY_TYPE` static, frozen attribute which gives us a hint to broadcast to all successive arrays.

kaushikcfd:
Closes #49.

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -375,15 +375,16 @@ def with_container_arithmetic(
                     gen(f"return cls({zip_init_args})")
 
                 if _bcast_actx_array_type:
-                    ary_types: Tuple[str, ...] = ("*arg1.array_context.array_types",)
+                    bcast_actx_ary_types: Tuple[str, ...] = (
+                        "*arg1.array_context.array_types",)
                 else:
-                    ary_types = ()
+                    bcast_actx_ary_types = ()
 
                 gen(f"""
                 if {bool(outer_bcast_type_names)}:  # optimized away
                     if isinstance(arg2,
                                   {tup_str(outer_bcast_type_names
-                                           + ary_types)}):
+                                           + bcast_actx_ary_types)}):
                         return cls({bcast_same_cls_init_args})
                 if {numpy_pred("arg2")}:
                     result = np.empty_like(arg2, dtype=object)
@@ -409,9 +410,9 @@ def with_container_arithmetic(
                         })
 
                 if _bcast_actx_array_type:
-                    ary_types = ("*arg2.array_context.array_types",)
+                    bcast_actx_ary_types = ("*arg2.array_context.array_types",)
                 else:
-                    ary_types = ()
+                    bcast_actx_ary_types = ()
 
                 gen(f"""
                     def {fname}(arg2, arg1):
@@ -420,7 +421,7 @@ def with_container_arithmetic(
                         if {bool(outer_bcast_type_names)}:  # optimized away
                             if isinstance(arg1,
                                           {tup_str(outer_bcast_type_names
-                                                   + ary_types)}):
+                                                   + bcast_actx_ary_types)}):
                                 return cls({bcast_init_args})
                         if {numpy_pred("arg1")}:
                             result = np.empty_like(arg1, dtype=object)
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -127,7 +127,7 @@ def _format_binary_op_str(op_str: str,
 def with_container_arithmetic(
         *,
         bcast_number: bool = True,
-        bcast_actx_array_type: Optional[bool] = None,
+        _bcast_actx_array_type: Optional[bool] = None,
         bcast_obj_array: Optional[bool] = None,
         bcast_numpy_array: bool = False,
         bcast_container_types: Optional[Tuple[type, ...]] = None,
@@ -143,7 +143,7 @@ def with_container_arithmetic(
 
     :arg bcast_number: If *True*, numbers broadcast over the container
         (with the container as the 'outer' structure).
-    :arg bcast_actx_array_type: If *True*, instances of base array types of the
+    :arg _bcast_actx_array_type: If *True*, instances of base array types of the
         container's array context are broadcasted over the container. Can be
         *True* only if the container has *_cls_has_array_context_attr* set.
         Defaulted to *bcast_number* if *_cls_has_array_context_attr* is set,
@@ -215,14 +215,14 @@ def with_container_arithmetic(
     if not bcast_obj_array and bcast_numpy_array:
         raise TypeError("bcast_obj_array must be set if bcast_numpy_array is")
 
-    if bcast_actx_array_type is None:
+    if _bcast_actx_array_type is None:
         if _cls_has_array_context_attr:
-            bcast_actx_array_type = bcast_number
+            _bcast_actx_array_type = bcast_number
         else:
-            bcast_actx_array_type = False
+            _bcast_actx_array_type = False
     else:
-        if bcast_actx_array_type and not _cls_has_array_context_attr:
-            raise TypeError("bcast_actx_array_type can be True iff "
+        if _bcast_actx_array_type and not _cls_has_array_context_attr:
+            raise TypeError("_bcast_actx_array_type can be True only if "
                             "_cls_has_array_context_attr is set.")
 
     if bcast_numpy_array:
@@ -374,7 +374,7 @@ def with_container_arithmetic(
                                     raise ValueError(msg)""")
                     gen(f"return cls({zip_init_args})")
 
-                if bcast_actx_array_type:
+                if _bcast_actx_array_type:
                     all_outer_bcast_type_names = (
                         outer_bcast_type_names
                         + ("*arg1.array_context.array_types",))
@@ -400,7 +400,7 @@ def with_container_arithmetic(
             # {{{ "reverse" binary operators
 
             if reversible:
-                if bcast_actx_array_type:
+                if _bcast_actx_array_type:
                     all_outer_bcast_type_names = (
                         outer_bcast_type_names
                         + ("*arg2.array_context.array_types",))
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -375,15 +375,15 @@ def with_container_arithmetic(
                     gen(f"return cls({zip_init_args})")
 
                 if _bcast_actx_array_type:
-                    all_outer_bcast_type_names = (
-                        outer_bcast_type_names
-                        + ("*arg1.array_context.array_types",))
+                    ary_types: Tuple[str, ...] = ("*arg1.array_context.array_types",)
                 else:
-                    all_outer_bcast_type_names = outer_bcast_type_names
+                    ary_types = ()
 
                 gen(f"""
                 if {bool(outer_bcast_type_names)}:  # optimized away
-                    if isinstance(arg2, {tup_str(all_outer_bcast_type_names)}):
+                    if isinstance(arg2,
+                                  {tup_str(outer_bcast_type_names
+                                           + ary_types)}):
                         return cls({bcast_same_cls_init_args})
                 if {numpy_pred("arg2")}:
                     result = np.empty_like(arg2, dtype=object)
@@ -400,12 +400,6 @@ def with_container_arithmetic(
             # {{{ "reverse" binary operators
 
             if reversible:
-                if _bcast_actx_array_type:
-                    all_outer_bcast_type_names = (
-                        outer_bcast_type_names
-                        + ("*arg2.array_context.array_types",))
-                else:
-                    all_outer_bcast_type_names = outer_bcast_type_names
                 fname = f"_{cls.__name__.lower()}_r{dunder_name}"
                 bcast_init_args = cls._deserialize_init_arrays_code("arg2", {
                         key_arg2: _format_binary_op_str(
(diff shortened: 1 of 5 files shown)


alexfikl:
Is the "not device scalar" case useful?

In the general case, `DOFArray` will have different groups with different numbers of elements and degrees of freedom, so doing `DOFArray + my_cl_array` is unlikely to work, right?

kaushikcfd:
>In the general case, DOFArray will have different groups with different numbers of elements and degrees of freedom, so doing DOFArray + my_cl_array is unlikely to work, right?

If the broadcasting for any of the leaf arrays is illegal, we would see an error, which IMO is a reasonable user experience.

arraycontext/container/arithmetic.py
```diff
@@ -343,6 +343,17 @@ def {fname}(arg1):
...
                     gen(f"return cls({zip_init_args})")
+                if _cls_has_array_context_attr:
+                    gen("if isinstance(arg2,"
+                        "              arg1.array_context.get_array_types()):")
+                    with Indentation(gen):
+                        gen(f"""
+                                return (
+                                    rec_map_array_container(
+                                        lambda a: {op_str.format("a", "arg2")},
+                                        arg1))
```
alexfikl, line 354: This should use those `_deserialize_init_arrays_code` and `_serialize_init_arrays_code` methods to unwrap the serialization. There should be an example for numpy object arrays lower down (?).

arraycontext/context.py
```diff
@@ -157,6 +157,13 @@ def _get_fake_numpy_namespace(self):
         from .fake_numpy import BaseFakeNumpyNamespace
         return BaseFakeNumpyNamespace(self)
 
+    def get_array_types(self):
+        """
+        Returns a :class:`tuple` of types that are valid leaf array classes in
+        the context.
```
alexfikl, line 163:
```suggestion
        Returns a :class:`tuple` of types that are valid base array classes
        the context can operate on.
```
?

arraycontext/container/arithmetic.py
```diff
@@ -343,6 +343,17 @@ def {fname}(arg1):
                                 else:
                                     raise ValueError(msg)""")
                     gen(f"return cls({zip_init_args})")
+                if _cls_has_array_context_attr:
```
inducer, line 346: This should get its own flag, defaulted to the same value as `bcast_number`.

arraycontext/container/arithmetic.py
```diff
@@ -343,6 +343,17 @@ def {fname}(arg1):
...
+                    gen("if isinstance(arg2,"
```
inducer, line 347: Ordering here matters a great deal. (Maybe there should be a comment stating this.) These cases should be sorted from most likely to least. Is this the second-most-likely case?

arraycontext/container/arithmetic.py
```diff
@@ -343,6 +343,17 @@ def {fname}(arg1):
...
+                        gen(f"""
+                                return (
+                                    rec_map_array_container(
+                                        lambda a: {op_str.format("a", "arg2")},
+                                        arg1))
+                            """)
+
                 gen(f"""
                 if {bool(outer_bcast_type_names)}:  # optimized away
                     if isinstance(arg2, {tup_str(outer_bcast_type_names)}):
```
inducer, line 359: This should also work (and be tested) for the reverse operators.

arraycontext/container/arithmetic.py
```diff
@@ -343,6 +343,17 @@ def {fname}(arg1):
                                 else:
                                     raise ValueError(msg)""")
                     gen(f"return cls({zip_init_args})")
+                if _cls_has_array_context_attr:
+                    gen("if isinstance(arg2,"
+                        "              arg1.array_context.get_array_types()):")
```
inducer, line 348: (I can't seem to comment on the same line multiple times, hence more comments on subsequent lines.) `isinstance` is rather expensive. Can we get by with `arg2.__class__ is ...`?

inducer (commented):
Thanks @alexfikl for taking a look. A few more thoughts below.

arraycontext/container/arithmetic.py
kaushikcfd: That would be pretty problematic as far as pytato is concerned. There isn't one single array class.

inducer: Good point.

arraycontext/context.py
kaushikcfd: Thanks, 4790d86.

arraycontext/container/arithmetic.py
kaushikcfd: I rather just used appended the types in broadcasting logic for numbers.

arraycontext/container/arithmetic.py
kaushikcfd: Done, see here: 93013c9.

arraycontext/container/arithmetic.py
kaushikcfd: IMO, this use-case would come just after host-scalars, as implemented in  93013c9.

arraycontext/container/arithmetic.py
kaushikcfd: Added a test in  f0e0587.

arraycontext/container/arithmetic.py
```diff
@@ -142,6 +143,11 @@ def with_container_arithmetic(
 
     :arg bcast_number: If *True*, numbers broadcast over the container
         (with the container as the 'outer' structure).
+    :arg bcast_actx_array_type: If *True*, instances of base array types of the
+        container's array context are broadcasted over the container. Can be
+        *True* iff the container has *_cls_has_array_context_attr* set.
```
alexfikl, line 148:
```suggestion
        *True* only if the container has *_cls_has_array_context_attr* set.
```
? To avoid the (technical?) contraction.

alexfikl (approved):
Looks good to me!

Just left a very small nitpick.

arraycontext/container/arithmetic.py
kaushikcfd: Sure, thanks!

arraycontext/container/arithmetic.py
```diff
@@ -357,10 +373,18 @@ def {fname}(arg1):
                                 else:
                                     raise ValueError(msg)""")
                     gen(f"return cls({zip_init_args})")
+
+                if bcast_actx_array_type:
+                    all_outer_bcast_type_names = (
+                        outer_bcast_type_names
+                        + ("*arg1.array_context.get_array_types()",))
+                else:
+                    all_outer_bcast_type_names = outer_bcast_type_names
```
inducer, line 382: * Move this closer to the usage site.
* Make this produce a tuple that's either empty or has the actx array type(s) in it. Add both together in the argument of `tup_str`.

(Same below for reverse.)

inducer (commented):
Thanks! Some comments below.

inducer:

test/test_arraycontext.py
```diff
@@ -877,6 +877,36 @@ def test_abs_complex(actx_factory):
...
+@dataclass(frozen=True)
+class Foo:
+    u: DOFArray
+
+    @property
+    def array_context(self):
+        return self.u.array_context
+
+
+def test_leaf_array_type_broadcasting(actx_factory):
```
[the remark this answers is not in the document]
kaushikcfd: Done  974c242

test/test_arraycontext.py
```diff
@@ -877,6 +877,36 @@ def test_abs_complex(actx_factory):
...
+    actx = actx_factory()
+
+    foo = Foo(DOFArray(actx, (actx.zeros(3, dtype=np.float64) + 41, )))
+    bar = foo + 4
+    baz = foo + actx.from_numpy(4*np.ones((3, )))
+    qux = actx.from_numpy(4*np.ones((3, ))) + foo
```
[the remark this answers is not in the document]
kaushikcfd: Done  974c242.

arraycontext/context.py
```diff
@@ -158,6 +158,13 @@ def _get_fake_numpy_namespace(self):
         from .fake_numpy import BaseFakeNumpyNamespace
         return BaseFakeNumpyNamespace(self)
 
+    def get_array_types(self):
+        """
+        Returns a :class:`tuple` of types that are valid base array classes
+        the context can operate on.
+        """
+        return ()
```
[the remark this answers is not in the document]
kaushikcfd: Done.  9255f22

> Would it be sensible to only allow one type here, to avoid the splat above?

There is some value in keeping it tuple, default value is much nicer since the only intent is to perform type checking using it.

arraycontext/container/arithmetic.py
```diff
@@ -142,6 +143,11 @@ def with_container_arithmetic(
 
     :arg bcast_number: If *True*, numbers broadcast over the container
         (with the container as the 'outer' structure).
+    :arg bcast_actx_array_type: If *True*, instances of base array types of the
```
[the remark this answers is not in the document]
kaushikcfd: Done 96392c0.

arraycontext/container/arithmetic.py
```diff
@@ -209,6 +215,16 @@ def _deserialize_init_arrays_code(cls, tmpl_instance_name, args):
...
         raise TypeError("bcast_obj_array must be set if bcast_numpy_array is")
 
+    if bcast_actx_array_type is None:
+        if _cls_has_array_context_attr:
+            bcast_actx_array_type = bcast_number
+        else:
+            bcast_actx_array_type = False
+    else:
+        if bcast_actx_array_type and not _cls_has_array_context_attr:
+            raise TypeError("bcast_actx_array_type can be True iff "
```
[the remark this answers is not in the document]
kaushikcfd: Done  96392c0.

arraycontext/container/arithmetic.py
kaushikcfd: Thanks, I think 2040a8d should fix it.

arraycontext/container/arithmetic.py
```diff
@@ -357,10 +373,18 @@ def {fname}(arg1):
                                 else:
                                     raise ValueError(msg)""")
                     gen(f"return cls({zip_init_args})")
+
+                if _bcast_actx_array_type:
+                    ary_types: Tuple[str, ...] = ("*arg1.array_context.array_types",)
```
inducer, line 378:
```suggestion
                    bcast_actx_ary_types: Tuple[str, ...] = ("*arg1.array_context.array_types",)
```

arraycontext/container/arithmetic.py
```diff
@@ -383,12 +407,20 @@ def {fname}(arg1):
                         for key_arg2, expr_arg2 in
                         cls._serialize_init_arrays_code("arg2").items()
                         })
+
+                if _bcast_actx_array_type:
+                    ary_types = ("*arg2.array_context.array_types",)
```
inducer, line 412:
```suggestion
                    bcast_actx_ary_types = ("*arg2.array_context.array_types",)
```

inducer (commented):
Thanks! Just a few more minor things, then this is ready to go.

inducer:

arraycontext/context.py
```diff
@@ -153,6 +158,7 @@ class ArrayContext(ABC):
 
     def __init__(self):
         self.np = self._get_fake_numpy_namespace()
+        self.array_types = ()
```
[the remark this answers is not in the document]
kaushikcfd: Agreed, e6b8b1b. Had to go with making it a property to avoid global import. Please if you foresee something wrong with this.

arraycontext/context.py
```diff
@@ -144,6 +144,11 @@ class ArrayContext(ABC):
          Callables accessible through this namespace vectorize over object
          arrays, including :class:`arraycontext.ArrayContainer`\ s.
 
+    .. attribute:: array_types
+
+        A :class:`tuple` of types that are valid base array classes
+        the context can operate on.
```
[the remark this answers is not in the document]
kaushikcfd: Thanks, eabe193.

arraycontext/container/arithmetic.py
kaushikcfd: 9eae9a9

arraycontext/container/arithmetic.py
kaushikcfd: 9eae9a9

arraycontext/context.py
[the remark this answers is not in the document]
inducer: What I meant was: set it was a class attribute here, but override it as an instance attribute in the subclasses. (Or does mypy not like that?) I definitely don't like the `@property` way much: it brings back all that function call overhead, and more. Tried that in e6b8b1b and 8702db6. Let me know if you hate that.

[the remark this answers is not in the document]
kaushikcfd: > it brings back all that function call overhead, and more. 

Aah I see. yep!

I like this, thanks!

PR #52, opened 2021-06-28
Test, fix equality-comparison behavior

### LINKED ISSUE #48: Make sure array containers handle __eq__ reasonably
A risk I perceive is that the dataclass implements an `eq` which just goes, hey, let's equality-compare all members, then decides, incorrectly, that, yup, things are equal.

I don't *know* that it's mishandling stuff (as in, I haven't *observed* unsafe behavior). I'm writing this issue as a reminder to check (and write a test).

x-ref: https://github.com/inducer/arraycontext/pull/46/files#r659211749

inducer:
Closes #48.

Good thing I checked... turns out that there were multiple bugs here. First, the spurious-equality scenario from #48 was possible, though under somewhat contrived (and non-default) circumstances.

Next, I spotted that we were misusing `enum.auto` in a way that made all "operator classes" the same. (Yikes!)

This takes care of both of those and adds a test.

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -42,12 +42,12 @@ T = TypeVar("T")
 
 
 class _OpClass(enum.Enum):
-    ARITHMETIC = enum.auto
-    MATMUL = enum.auto
-    BITWISE = enum.auto
-    SHIFT = enum.auto
-    EQ_COMPARISON = enum.auto
-    REL_COMPARISON = enum.auto
+    ARITHMETIC = enum.auto()
+    MATMUL = enum.auto()
+    BITWISE = enum.auto()
+    SHIFT = enum.auto()
+    EQ_COMPARISON = enum.auto()
+    REL_COMPARISON = enum.auto()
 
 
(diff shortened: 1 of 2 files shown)


arraycontext/container/arithmetic.py
```diff
@@ -42,12 +42,12 @@
 
 
 class _OpClass(enum.Enum):
-    ARITHMETIC = enum.auto
-    MATMUL = enum.auto
-    BITWISE = enum.auto
-    SHIFT = enum.auto
-    EQ_COMPARISON = enum.auto
-    REL_COMPARISON = enum.auto
+    ARITHMETIC = enum.auto()
```
alexfikl, line 45: :facepalm: 

https://docs.python.org/3/library/enum.html#enum.unique?

alexfikl (approved):
Well, this was a bit nasty. Nice catch!

arraycontext/container/arithmetic.py
inducer: Nice :adhesive_bandage:, TIL!

fc3b701

PR #53, opened 2021-06-30
Remove rec_multimap_array_container usage in actx.np.equal

alexfikl:
The previous behaviour seems incorrect, since it will recurse and compare even if the inputs don't have `@with_container_arithmetic(eq_comparison=True, rel_comparison=True)`.

- [ ] Record in the docs that equality comparison of array context arrays is not well-defined.

### DIFF
--- a/arraycontext/impl/pyopencl/fake_numpy.py
+++ b/arraycontext/impl/pyopencl/fake_numpy.py
@@ -27,7 +27,6 @@ THE SOFTWARE.
 """
 
 from functools import partial, reduce
-import operator
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


inducer:
I think the existing beahvior is justifiable, particularly if you look at the comment right above there:
https://github.com/inducer/arraycontext/blob/ec7d66716276fb64662f576252cdef47adebf6e4/arraycontext/impl/pyopencl/fake_numpy.py#L51-L56

What's the damage if these routines recurse (and are defined to do so)?

alexfikl:
> What's the damage if these routines recurse (and are defined to do so)?

They'll recurse with the proposed change as well if the array container has the proper arithmetic enabled. Of course that doesn't work for object arrays, so this is probably a no-go as is.

I guess this was more of a question than a PR: 
* Should these comparisons work for array containers with disabled comparisons?
* Should they broadcast in any way?

inducer:
> Should these comparisons work for array containers with disabled comparisons?

IMO yes. It helps to consider equality comparison and "relative" comparisons separately, IMO. Equality comparison is not well-defined on array context arrays globally: numpy-like ones will do the element-wise thing, while pytato ones will do the "yes/no" structural equality thing. So dispatching to it is asking for trouble. That's why the equality part of this is not popular with me. (Meanwhile, having a reliable way to ask for numpy-style equality comparisons is useful.)

For relative comparisons, both these functions and the operators will do the same thing, so the distinction is less clear. Should they work on types that have `rel_comparison` disabled? I don't see a reason why not, but I also don't *really* see a reason why anyone might disable `rel_comparison`. I'd be OK with checking and yelling if it's disabled for some reason.

> Should they broadcast in any way?

Since they represent a way to reliably ask for the numpy-like behavior, yes.

PR #55, opened 2021-06-30
Refactor actx elwise funcs

inducer:
Supersedes #54, with a more unified approach. Also moving some code to a more sensible place.

Nabbed unit test from there. (thanks @alexfikl!)


alexfikl (approved):
Looks much better! Thanks!

PR #56, opened 2021-07-01
Cache codegen result in freeze()

matthiasdiener:
Needs:
- [x] https://github.com/inducer/pytato/pull/118

### DIFF
--- a/arraycontext/impl/pytato/utils.py
+++ b/arraycontext/impl/pytato/utils.py
@@ -1,0 +1,82 @@
+__copyright__ = """
+Copyright (C) 2021 University of Illinois Board of Trustees
+"""
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+
+from typing import Any, Dict, Set, Tuple, Mapping
+from pytato.array import SizeParam, Placeholder
+from pytato.array import Array, DataWrapper
+from pytato.transform import CopyMapper
+from pytools import UniqueNameGenerator
+
+
+class _DatawrapperToBoundPlaceholderMapper(CopyMapper):
+    """
+    Helper mapper for :func:`normalize_pt_expr`. Every
+    :class:`pytato.DataWrapper` is replaced with a deterministic copy of
+    :class:`Placeholder`.
+    """
+    def __init__(self) -> None:
+        super().__init__()
+        self.bound_arguments: Dict[str, Any] = {}
+        self.vng = UniqueNameGenerator()
+        self.seen_inputs: Set[str] = set()
+
+    def map_data_wrapper(self, expr: DataWrapper) -> Array:
+        if expr.name is not None:
+            if expr.name in self.seen_inputs:
+                raise ValueError("Got multiple inputs with the name"
+                                 f"{expr.name} => Illegal.")
+            self.seen_inputs.add(expr.name)
+
+        # Normalizing names so that we more arrays can have the normalized DAG.
+        name = self.vng("_actx_dw")
+        self.bound_arguments[name] = expr.data
+        return Placeholder(name=name,
+                           shape=tuple(self.rec(s) if isinstance(s, Array) else s
+                                       for s in expr.shape),
+                           dtype=expr.dtype,
+                           tags=expr.tags)
+
+    def map_size_param(self, expr: SizeParam) -> Array:
+        raise NotImplementedError
+
+    def map_placeholder(self, expr: Placeholder) -> Array:
+        raise ValueError("Placeholders cannot appear in"
+                         " DatawrapperToBoundPlaceholderMapper.")
+
+
+def _normalize_pt_expr(expr: Array) -> Tuple[Array,
+                                            Mapping[str, Any]]:
+    """
+    Returns ``(normalized_expr, bound_arguments)``.  *normalized_expr* is a
+    normalized form of *expr*, with all instances of
+    :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
+    named in a deterministic manner. The data corresponding to the placeholders
+    in *normalized_expr* is recorded in the mapping *bound_arguments*.
+    Deterministic naming of placeholders permits more effective hashing of
+    equivalent graphs.
+    """
+    normalize_mapper = _DatawrapperToBoundPlaceholderMapper()
+    normalized_expr = normalize_mapper(expr)
+    return normalized_expr, normalize_mapper.bound_arguments
--- a/arraycontext/impl/pytato/utils.py
+++ b/arraycontext/impl/pytato/utils.py
@@ -74,7 +74,7 @@ def _normalize_pt_expr(expr: Array) -> Tuple[Array,
     :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
     named in a deterministic manner. The data corresponding to the placeholders
     in *normalized_expr* is recorded in the mapping *bound_arguments*.
-    Deterministic naming of placeholders permits more effective hashing of
+    Deterministic naming of placeholders permits more effective caching of
     equivalent graphs.
     """
     normalize_mapper = _DatawrapperToBoundPlaceholderMapper()
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -70,6 +70,7 @@ class PytatoPyOpenCLArrayContext(ArrayContext):
         self.queue = queue
         self.allocator = allocator
         self.array_types = (pt.Array, )
+        self._freeze_prg_cache = {}
 
         # unused, but necessary to keep the context alive
         self.context = self.queue.context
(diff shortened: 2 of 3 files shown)


inducer:
(Still needs a conflict resolution)

kaushikcfd:
Rebased on top of main, this is ready for a look. Thanks in advance!

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+def normalize_pt_expr(expr: Array) -> Tuple[Array,
+                                            Mapping[str, Any]]:
+    """
+    Returns ``(normalized_expr, bound_arguments)``.  *normalized_expr* is a
+    normalized form of *expr*, with all instances of
+    :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
+    in a deterministic manner. The data corresponding to the placeholders in
+    *normalized_expr* is recorded in the mapping *bound_arguments*.
+    """
```
matthiasdiener, line 114: Should this function be part of the documentation?

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                           shape=tuple(self.rec(s) if isinstance(s, Array) else s
+                                       for s in expr.shape),
+                           dtype=expr.dtype,
+                           tags=expr.tags)
+
+    def map_size_param(self, expr: BaseSizeParam) -> Array:
+        if expr.name in self.seen_inputs:
+            raise ValueError("Got multiple inputs with the name"
+                             f"{expr.name} => Illegal.")
+        assert isinstance(expr.name, str)
```
matthiasdiener, line 96: Out of curiosity, could this ever be not a string?

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __eq__(self, other) -> bool:
+        return Array.__eq__(self, other)
+
+    def __hash__(self) -> int:
+        return Array.__hash__(self)
+
+
+class Normalizer(CopyMapper):
+    def __init__(self) -> None:
```
matthiasdiener, line 71: Could add a brief doc here

matthiasdiener (commented):
LGTM, thanks! Just a few comments

arraycontext/impl/pytato/utils.py
kaushikcfd: Done, thanks!

arraycontext/impl/pytato/utils.py
kaushikcfd: This actually got me into thinking that handling `SizeParam` would be more than what we need to handle. raised a `NotImplementedError` for these for now.

arraycontext/impl/pytato/utils.py
kaushikcfd: I wouldn't expect these functions to be useful in the user-land, so I think it's fine if we don't publicly document them.

matthiasdiener:
Ready for review @inducer

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,97 @@
...
+                 dtype: np.dtype,
+                 tags: TagsType = frozenset()
+                 ) -> None:
+        super().__init__(name, shape, dtype, tags)
+
+    def __eq__(self, other) -> bool:
+        return Array.__eq__(self, other)
+
+    def __hash__(self) -> int:
+        return Array.__hash__(self)
```
inducer, line 49: Could you explain this? Isn't the base `PlaceHolder` already equality-comparable?

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        self.seen_inputs: Set[str] = set()
+
+    def map_data_wrapper(self, expr: DataWrapper) -> Array:
+        if expr.name is not None:
+            if expr.name in self.seen_inputs:
+                raise ValueError("Got multiple inputs with the name"
+                                 f"{expr.name} => Illegal.")
+            self.seen_inputs.add(expr.name)
+
+        name = self.vng("_actx_dw")
```
inducer, line 71: * What guarantees the uniqueness of these names?
* If you're given a name (via `expr.name`), why not use it?

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(name, shape, dtype, tags)
+
+    def __eq__(self, other) -> bool:
+        return Array.__eq__(self, other)
+
+    def __hash__(self) -> int:
+        return Array.__hash__(self)
+
+
+class Normalizer(CopyMapper):
```
inducer, line 52:
```suggestion
class DatawrapperToBoundPlaceholderMapper(CopyMapper):
```

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    Returns ``(normalized_expr, bound_arguments)``.  *normalized_expr* is a
+    normalized form of *expr*, with all instances of
+    :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
+    in a deterministic manner. The data corresponding to the placeholders in
+    *normalized_expr* is recorded in the mapping *bound_arguments*.
+    """
+    normalize_mapper = Normalizer()
+    normalized_expr = normalize_mapper(expr)
+    return normalized_expr, normalize_mapper.bound_arguments
```
inducer, line 97: Is this wrapper really necessary?

inducer (commented):
Thanks! A few things below.

inducer:

arraycontext/impl/pytato/utils.py
kaushikcfd: But not in a persistent manner. The equality checks are essentially identity checks https://github.com/inducer/pytato/blob/8f6a07fe041bc5a4bf35e770005f1b25f9b2efb0/pytato/array.py#L1436-L1440.

arraycontext/impl/pytato/utils.py
kaushikcfd: > What guarantees the uniqueness of these names?

We use a `pytools.UniqueNameGenerator` to get these names, so must be unique.

> If you're given a name (via expr.name), why not use it?

More pairs of isomorphic DAGs normalized to the same DAG.

arraycontext/impl/pytato/utils.py
kaushikcfd: Yep, see https://github.com/inducer/arraycontext/pull/56#discussion_r669565091. Otherwise we can (almost) never have 2 isomorphic DAGs being being equal.

arraycontext/impl/pytato/utils.py
kaushikcfd: Thanks, done!

arraycontext/impl/pytato/utils.py
inducer: OIC. Question back at you then: Why did we make that choice in `InputArgumentBase`? It seems counterproductive. For a `DataWrapper`, maybe I could see that... but `Placeholder` and such shouldn't really do that.

arraycontext/impl/pytato/utils.py
inducer: > so must be unique.

I guess my question was "among what", and the answer I'm hearing is "only among themselves". Could you document that? (And that's OK because you're banning other named things.)

> More pairs of isomorphic DAGs normalized to the same DAG at a very minor cost.

Hmm... I see. Add a comment?

arraycontext/impl/pytato/utils.py
inducer: What I meant is not "we don't need this functionality", but rather "people can call the mapper directly".

arraycontext/impl/pytato/utils.py
kaushikcfd: Although I agree that making Placeholders equality comparable seems to be the right thing to do, there are some practical limitations to it. As, all our mappers are "cached" and there would be no way practically feasible way to determine illegal DAGs like
```python
a = pt,make_placeholder("foo", (10, 4))
b = pt,make_placeholder("foo", (10, 4)) 
my_illegal_dag = a + b # unidentical placeholders with same name is most-likely a user-error that should be caught
```

kaushikcfd: > practically feasible way

I think caching the results with `id` for a mapper isn't too bad, we can do this. I don't see any big problems with this.

kaushikcfd: Since, this is more pytato-specific, I vote to move this discussion there. See https://github.com/inducer/pytato/pull/123. Please un-resolved if anyone disagrees.

inducer: Given that https://github.com/inducer/pytato/pull/123 is now (as good as) in, could this be deleted?

arraycontext/impl/pytato/utils.py
inducer: Could you add a comment motivating the name normalization?

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,98 @@
...
+                         " DatawrapperToBoundPlaceholderMapper.")
+
+
+def normalize_pt_expr(expr: Array) -> Tuple[Array,
+                                            Mapping[str, Any]]:
+    """
+    Returns ``(normalized_expr, bound_arguments)``.  *normalized_expr* is a
+    normalized form of *expr*, with all instances of
+    :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
+    in a deterministic manner. The data corresponding to the placeholders in
```
inducer, line 93:
```suggestion
    named in a deterministic manner. The data corresponding to the placeholders in
```

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    *normalized_expr* is recorded in the mapping *bound_arguments*.
```
inducer, line 94:
```suggestion
    *normalized_expr* is recorded in the mapping *bound_arguments*.
    Deterministic naming of placeholders permits more effective hashing of
    equivalent graphs.
```

kaushikcfd: Done.

arraycontext/impl/pytato/utils.py
kaushikcfd: Done.

arraycontext/impl/pytato/utils.py
kaushikcfd: Added.

arraycontext/impl/pytato/utils.py
kaushikcfd: Yep, removed.

arraycontext/impl/pytato/utils.py
kaushikcfd: Adding those three lines to `PytatoPyOpenCLArrayContext.freeze` was looking a bit out-of-character. Made the routine an underscored one.

arraycontext/impl/pytato/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+def _normalize_pt_expr(expr: Array) -> Tuple[Array,
+                                            Mapping[str, Any]]:
+    """
+    Returns ``(normalized_expr, bound_arguments)``.  *normalized_expr* is a
+    normalized form of *expr*, with all instances of
+    :class:`pytato.DataWrapper` replaced with instances of :class:`Placeholder`
+    named in a deterministic manner. The data corresponding to the placeholders
+    in *normalized_expr* is recorded in the mapping *bound_arguments*.
+    Deterministic naming of placeholders permits more effective hashing of
```
inducer, line 77:
```suggestion
    Deterministic naming of placeholders permits more effective caching of
```

inducer (approved):
Thanks! I'll apply the one suggestion. This is ready to go in.

PR #58, opened 2021-07-06
call_loopy: correct entrypoint

matthiasdiener:


kaushikcfd (approved):
Thanks, I agree, this should use `default_entrypoint`. Since the user would be passing an executable translation unit without any entrypoint info, the only kernel that would be executable would be the only entrypoint present.

PR #62, opened 2021-07-15
Add generic container reductions

### LINKED ISSUE #61: Make a utility for reduction over array container structures
cc @alexfikl 

Based on https://github.com/inducer/arraycontext/pull/60#discussion_r669042558

alexfikl:
This adds some helpers to do reductions over array containers.

* adds a generic `rec_reduce_array_container(reduce_func, array_func, ary)` function.
* uses that in `actx.np.[sum, min, max]`.

Fixes #61.

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -8,6 +8,9 @@
 .. autofunction:: rec_map_array_container
 .. autofunction:: rec_multimap_array_container
 
+.. autofunction:: rec_reduce_array_container
+.. autofunction:: rec_multireduce_array_container
+
 Traversing decorators
 ~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: mapped_over_array_containers
@@ -78,7 +81,7 @@ def _map_array_container_impl(
         if type(_ary) is leaf_cls:  # type(ary) is never None
             return f(_ary)
         elif is_array_container(_ary):
-            return make_container(_ary, [
+            return make_container(_ary, [           # type: ignore[operator]
                 (key, frec(subary)) for key, subary in serialize_container(_ary)
                 ])
         else:
@@ -93,7 +96,7 @@ def _map_array_container_impl(
 def _multimap_array_container_impl(
         f: Callable[..., Any],
         *args: Any,
-        reduce_func: Callable[[Any, Iterable[Any]], Any] = None,
+        reduce_func: Callable[[Any, Iterable[Tuple[Any, Any]]], Any] = None,
         leaf_cls: Optional[type] = None,
         recursive: bool = False) -> ArrayContainerT:
     """Helper for :func:`rec_multimap_array_container`.
@@ -128,9 +131,9 @@ def _multimap_array_container_impl(
 
                 new_args[i] = subary
 
-            result.append((key, frec(*new_args)))       # type: ignore
+            result.append((key, frec(*new_args)))       # type: ignore[operator]
 
-        return make_container(template_ary, result)
+        return make_container(template_ary, result)     # type: ignore[operator]
 
     container_indices: List[int] = [
             i for i, arg in enumerate(args)
@@ -155,6 +158,7 @@ def _multimap_array_container_impl(
 
     make_container = deserialize_container if reduce_func is None else reduce_func
     frec = rec if recursive else f
+
     return rec(*args)
 
 # }}}
@@ -280,7 +284,7 @@ def rec_keyed_map_array_container(f: Callable[[Tuple[Any, ...], Any], Any],
             _ary: ArrayContainerT) -> ArrayContainerT:
         if is_array_container(_ary):
             return deserialize_container(_ary, [
-                    (key, rec(keys+(key,), subary))
+                    (key, rec(keys + (key,), subary))
                     for key, subary in serialize_container(_ary)
                     ])
         else:
@@ -316,7 +320,7 @@ def rec_reduce_array_container(
 
 def rec_multireduce_array_container(
         reduce_func: Callable[[Iterable[Any]], Any],
-        array_func: Callable[[Any], Any],
+        array_func: Callable[..., Any],
         *args: Any) -> Any:
     """Perform reductions over multiple array containers recursively.
 
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -332,7 +332,7 @@ def rec_multimap_reduce_array_container(
         array of the same type or a scalar.
     """
     # NOTE: this wrapper matches the signature of `deserialize_container`
-    # to make pluggin into `_multimap_array_container_impl` easier
+    # to make plugging into `_multimap_array_container_impl` easier
     def _reduce_wrapper(ary: Any, iterable: Iterable[Tuple[Any, Any]]) -> Any:
         return reduce_func([subary for _, subary in iterable])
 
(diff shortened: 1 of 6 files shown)


arraycontext/container/traversal.py
```diff
@@ -279,6 +291,52 @@ def rec(keys: Tuple[Union[str, int], ...],
...
+
+    :param reduce_func: callable used to reduce over the components of the
+        :class:`~arraycontext.ArrayContainer`.
+    :param array_func: callable used to reduce a single component of the
+        :class:`~arraycontext.ArrayContainer`. The callable takes arrays of
+        type :class:`arraycontext.ArrayContext.array_types` and returns an
+        array of the same type or a scalar.
+    """
+    def _reduce_wrapper(ary: Any, iterable: Iterable[Tuple[Any, Any]]) -> Any:
+        return reduce_func([subary for _, subary in iterable])
```
kaushikcfd, line 314: Keep this a generator to avoid memory allocation?

arraycontext/container/traversal.py
```diff
@@ -279,6 +291,52 @@ def rec(keys: Tuple[Union[str, int], ...],
 # }}}
 
 
+# {{{ array container reductions
+
+def rec_reduce_array_container(
```
kaushikcfd, line 300: Seems to me like it is performing 2 operations at once i.e. map_and_reduce. How about we only provide a primitive for performing reduction over the children of a node. The user could call `rec_map_array_container` and then chain the call to `rec_reduce_array_container` over the mapped sub-arrays.

kaushikcfd (commented):
Thanks! Left some high-level design choice questions.

arraycontext/container/traversal.py
```diff
@@ -279,6 +295,52 @@ def rec(keys: Tuple[Union[str, int], ...],
...
+    """Perform reductions over array containers recursively.
+
+    :param reduce_func: callable used to reduce over the components of the
+        :class:`~arraycontext.ArrayContainer`.
+    :param array_func: callable used to reduce a single component of the
+        :class:`~arraycontext.ArrayContainer`. The callable takes arrays of
+        type :class:`arraycontext.ArrayContext.array_types` and returns an
+        array of the same type or a scalar.
+    """
+    def _reduce_wrapper(ary: Any, iterable: Iterable[Tuple[Any, Any]]) -> Any:
```
inducer, line 313: Why not avoid this wrapper altogether? (And just change the prototype of `reduce_func` to match?)

arraycontext/container/traversal.py
```diff
@@ -59,11 +62,12 @@
         serialize_container, deserialize_container)
 
 
-# {{{ array container traversal
+# {{{ array container traversal helpers
 
 def _map_array_container_impl(
         f: Callable[[Any], Any],
         ary: ArrayContainerT, *,
+        reduce_func: Callable[[Any, Iterable[Tuple[Any, Any]]], Any] = None,
```
inducer, line 70: Honestly, not sure I'm loving this. (Writing this after reading much of the rest of this PR.) Why not make something specific to reductions instead of trying to generalize this by force? (via lots of inefficient function wrapping)

inducer (commented):
Thanks! A few comments below.

inducer:

arraycontext/container/traversal.py
alexfikl: @inducer The main reason I did it like this was because `_multimap_array_container_impl` has a lot more logic in there that I felt squimish about largely copy pasting for the reduction case. I can add an if in there to avoid the wrapping. What do you think?

inducer: I see your point. How about this: Duplicate `map`, but leave the wild wrap-and-dispatch thing in place for `multimap`? I can't really think of a "pretty" way to deal with this.

arraycontext/container/traversal.py
alexfikl: Hm, that's a good point. I would prefer to keep it (mostly) as is because
* We don't have any use case at the moment for reduce only, right?
* This avoid traversing the container twice to do a mapreduce.

I'll rename the functions to `rec_map_reduce_array_container` to be clearer that the `array_func` thing can be more general. What do you think?

alexfikl: 93f6594

arraycontext/container/traversal.py
```diff
@@ -77,19 +81,22 @@ def rec(_ary: ArrayContainerT) -> ArrayContainerT:
...
-            return deserialize_container(_ary, [
-                    (key, frec(subary)) for key, subary in serialize_container(_ary)
-                    ])
+            return make_container(_ary, [           # type: ignore[operator]
+                (key, frec(subary)) for key, subary in serialize_container(_ary)
+                ])
         else:
             return f(_ary)
 
+    make_container = deserialize_container if reduce_func is None else reduce_func
```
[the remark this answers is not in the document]
alexfikl: Done in 3989399.

arraycontext/container/traversal.py
alexfikl: Done in 94e7c6f.

arraycontext/container/traversal.py
alexfikl: Not really a fan because it pushes an implementation detail on the user. Although maybe there is some fancy reduction that would need to know about the `template_ary` and keys being reduced?

This is only used in the `multimap` case now. Do you still think it should be changed?

arraycontext/container/traversal.py
alexfikl: I'm on the side of keeping it a list too (only used in one place now too). How strongly do you feel about this?

arraycontext/container/traversal.py
inducer: Alright with me if it's `multimap`-only.

arraycontext/container/traversal.py
```diff
@@ -279,6 +292,57 @@ def rec(keys: Tuple[Union[str, int], ...],
...
+
+    :param reduce_func: callable used to reduce over the components of the
+        :class:`~arraycontext.ArrayContainer`.
+    :param map_func: callable used to map a single component of the
+        :class:`~arraycontext.ArrayContainer`. The callable takes arrays of
+        type :class:`arraycontext.ArrayContext.array_types` and returns an
+        array of the same type or a scalar.
+    """
+    # NOTE: this wrapper matches the signature of `deserialize_container`
+    # to make pluggin into `_multimap_array_container_impl` easier
```
inducer, line 335:
```suggestion
    # to make plugging into `_multimap_array_container_impl` easier
```

inducer (approved):
One typo, otherwise LGTM. Thanks for working on this! @kaushikcfd?

arraycontext/container/traversal.py
kaushikcfd: Yep, I think the name change makes it explicit. And this is fine by me.

arraycontext/container/traversal.py
kaushikcfd: Was a minor suggestion. But looks like generators have their downsides too, that was overlooked. Let's resolve it.

kaushikcfd (approved):
Thanks, this looks great to me! The same piece of logic was being re-used at multiple places and this provides a nice primitive for it.

inducer:
:tada: In it goes.

alexfikl:
> Thanks, this looks great to me! The same piece of logic was being re-used at multiple places and this provides a nice primitive for it.

Thanks for taking a look!

PR #63, opened 2021-07-16
bugfix: change order of pt.make_placeholder

kaushikcfd:
Sorry for the CI breakage on main.


kaushikcfd:
I'll set this to automerge. Seems pretty obvious and only way to fix it.

inducer:
Thanks!

PR #66, opened 2021-07-21
Add any and all to PyOpenCLArrayContext

alexfikl:


test/test_arraycontext.py
```diff
@@ -120,6 +120,14 @@ def __init__(self, actx, data):
 
     __array_priority__ = 10
 
+    def __bool__(self):
+        if len(self) == 1 and len(self.data[0]) == 1:
```
kaushikcfd, line 124: I think this checks whether the only group array is 1-long along its 0th-axis, but we might want to check if has a size of 1.

test/test_arraycontext.py
```diff
@@ -120,6 +120,14 @@ def __init__(self, actx, data):
...
+            return bool(self.data[0])
```
kaushikcfd, line 125: We should probably check the truth value of the to-numpy-ed scalar.

kaushikcfd (approved):
LGTM, thanks!

test/test_arraycontext.py
alexfikl: Good catch! Updated in 78dfbe6.

I added `__bool__` to meshmode in inducer/meshmode#251 too, but without the check for scalars. That forces a `to_numpy` essentially, so it seemed like a good idea to avoid it all together and just use this `any` / `all`. What do you think?

test/test_arraycontext.py
alexfikl: You mean add a test that actually exercises this?

kaushikcfd: > do self.array_context.to_numpy(self.data[0]) != 0?

I meant this. (sorry for the ambiguity)

test/test_arraycontext.py
kaushikcfd: Yep, I too am in favor of avoiding implicit `to_numpy`s.

test/test_arraycontext.py
kaushikcfd: Resolving since you already implemented it.

inducer (approved):

PR #67, opened 2021-07-29
PytatoPyOpenCLArrayContext.compile: add support for keyword arguments

kaushikcfd:


arraycontext/impl/pytato/compile.py
```diff
@@ -90,7 +90,8 @@ def _rec_str(key: Any) -> str:
     return "_".join(_rec_str(key) for key in keys)
 
 
-def _get_arg_id_to_arg_and_arg_id_to_descr(args: Tuple[Any, ...]
+def _get_arg_id_to_arg_and_arg_id_to_descr(args: Tuple[Any, ...],
+                                           kwargs: Mapping[Union[str], Any]
```
alexfikl, line 94: Why take a union of a single type here? Typo?

arraycontext/impl/pytato/compile.py
```diff
@@ -106,14 +107,15 @@ def _get_arg_id_to_arg_and_arg_id_to_descr(args: Tuple[Any, ...]
     arg_id_to_arg: Dict[Tuple[Any, ...], Any] = {}
     arg_id_to_descr: Dict[Tuple[Any, ...], AbstractInputDescriptor] = {}
 
-    for iarg, arg in enumerate(args):
+    for kw, arg in (tuple(enumerate(args))
+                    + tuple(kwargs.items())):
```
alexfikl, line 111: Use [itertools.chain](https://docs.python.org/3/library/itertools.html#itertools.chain) to avoid creating tuples?

alexfikl (approved):
Two small nitpicks, otherwise this looks good to me!

arraycontext/impl/pytato/compile.py
kaushikcfd: Whoops, was a leftover from  a previous version. Fixed as a part of  78ff8ff. Thanks for catching that!

arraycontext/impl/pytato/compile.py
kaushikcfd: Thanks! Done in  78ff8ff.

inducer:
Thanks!

PR #69, opened 2021-08-01
[PytatoPyOpenCLArrayContext] Do not evaluate array expressions for 0-sized arrays

kaushikcfd:


arraycontext/impl/pytato/__init__.py
```diff
@@ -123,6 +123,16 @@ def freeze(self, array):
             raise TypeError("PytatoPyOpenCLArrayContext.freeze invoked with "
                             f"non-pytato array of type '{type(array)}'")
 
+        # {{{ early exit for 0-sized arrays
+
+        if array.size == 0:
+            return cla.zeros(self.queue,
+                             shape=array.shape,
+                             dtype=array.dtype,
+                             allocator=self.allocator).with_queue(None)
```
inducer, line 132:
```suggestion
            return cla.empty(self.queue.context,
                             shape=array.shape,
                             dtype=array.dtype,
                             allocator=self.allocator)
```

inducer:
Thanks! (Made a small change, hope that's OK.)

PR #71, opened 2021-08-11
Add array_equal to PyOpenCLArrayContext.np

majosm:
Used by #46.

I'm not sure if there's a way to also make this work for `PytatoPyOpenCLArrayContext`? I ran into trouble with `DataWrapper`'s `__eq__` when I tried it.

### DIFF
--- a/arraycontext/impl/pyopencl/fake_numpy.py
+++ b/arraycontext/impl/pyopencl/fake_numpy.py
@@ -31,9 +31,13 @@ import operator
 
 from arraycontext.fake_numpy import \
         BaseFakeNumpyNamespace, BaseFakeNumpyLinalgNamespace
+from arraycontext.container import is_array_container
 from arraycontext.container.traversal import (
-        rec_multimap_array_container, rec_map_array_container,
+        rec_map_array_container,
+        rec_multimap_array_container,
+        multimap_reduce_array_container,
         rec_map_reduce_array_container,
+        rec_multimap_reduce_array_container,
         )
 
 try:
@@ -201,7 +205,6 @@ class PyOpenCLFakeNumpyNamespace(BaseFakeNumpyNamespace):
         return rec_map_array_container(_rec_ravel, a)
 
     def vdot(self, x, y, dtype=None):
-        from arraycontext import rec_multimap_reduce_array_container
         result = rec_multimap_reduce_array_container(
                 sum,
                 partial(cl_array.vdot, dtype=dtype, queue=self._array_context.queue),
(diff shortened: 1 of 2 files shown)


kaushikcfd:
> I ran into trouble with DataWrapper's __eq__ when I tried it.

For pytato arrays `Array.__eq__` is not compatiable with numpy. Maybe you want [pt.equal](https://documen.tician.de/pytato/array.html?highlight=equal#pytato.equal). But we don't have `(all|any)` implemented for pytato arrays as of yet.

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -208,6 +210,31 @@ def all(self, a):
...
+        elif not is_array_container(a):
+            same_shape = a.shape == b.shape
+            if same_shape:
+                result = (a == b).all()
+            else:
+                result = as_scalar(False)
+        else:
+            queue = self._array_context.queue
+            result = rec_multimap_reduce_array_container(
+                    partial(reduce, partial(cl_array.minimum, queue=queue)),
```
inducer, line 230: Is `minimum` documented to do the right thing applied to `bool` arrays?

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -208,6 +210,31 @@ def all(self, a):
             result = result.get()[()]
         return result
 
+    def array_equal(self, a, b):
+        def as_scalar(bool_value):
```
inducer, line 214: This seems to unnecessarily round-trip to the device. Why not just return the (known) `bool` result?

inducer (commented):
Thanks! A few notes below.

inducer:

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -208,6 +210,31 @@ def all(self, a):
             result = result.get()[()]
         return result
 
+    def array_equal(self, a, b):
```
inducer, line 213: @kaushikcfd What do you think this should return? Device scalar or "just plain `bool`"?

alexfikl: For what it's worth, the other reductions (e.g. `any`) seem to return a device scalar if given arrays and a host scalar if given host scalars.

kaushikcfd: I agree with @alexfikl, we should probably avoid implicit freeze-ing as much as possible.

arraycontext/impl/pyopencl/fake_numpy.py
majosm: I separated the device-to-host conversion from the recursion step, so I think this should only be operating on `int8` arrays now.

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -208,6 +211,31 @@ def all(self, a):
...
+        def rec_equal(x, y):
+            if type(x) != type(y):
+                return as_device_scalar(False)
+            elif not is_array_container(x):
+                return as_device_scalar(x.shape == y.shape) and (x == y).all()
+            else:
+                queue = self._array_context.queue
+                return multimap_reduce_array_container(
+                    partial(reduce, partial(cl_array.minimum, queue=queue)),
+                    rec_equal, x, y)
```
inducer, line 231:
```suggestion
                    # reduce_func=
                    partial(reduce, partial(cl_array.minimum, queue=queue)),
                    # map_func=
                    rec_equal,
                    x, y)
```

inducer (approved):

inducer:
Thx!

PR #75, opened 2021-08-17
PytatoPyOpenCLArrayContext: respect arraycontext.loopy's options

kaushikcfd:


arraycontext/impl/pytato/__init__.py
```diff
@@ -150,8 +150,7 @@ def freeze(self, array):
             pt_prg = self._freeze_prg_cache[normalized_expr]
         except KeyError:
             pt_prg = pt.generate_loopy(self.transform_dag(normalized_expr),
-                                       options=lp.Options(return_dict=True,
-                                                          no_numpy=True),
+                                       options=_DEFAULT_LOOPY_OPTIONS,
```
inducer, line 153: * I'd prefer if this could be customized per instance of the pytato array context. This is too global for my taste.
* Also, it should check that the important options are set.

PR #76, opened 2021-08-17
Allow actx.np.real on real arrays

alexfikl:
For some reason loopy (in [loopy_implemented_elwise_func](https://github.com/inducer/arraycontext/blob/35dc261453985a96df1cb05aa4d4c06d0b25eb9e/arraycontext/fake_numpy.py#L149-L155)) wasn't able to infer the output type of `real` when acting on real arrays. There's probably a better fix for this somewhere down the stack, but in the meantime.

For context, hit this while trying to use [pytential.norm](https://github.com/inducer/pytential/blob/a09934a89e00bb4bbbaebd435e01f02b2fa2e5e9/pytential/__init__.py#L77).


alexfikl:
Fixed by inducer/loopy#486 for pyopencl targets.

PR #77, opened 2021-08-24
use the standard doc config

matthiasdiener:
Taken from pytato.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,10 +14,6 @@ exec(compile(open("../arraycontext/version.py").read(), "../arraycontext/version
 version = ".".join(str(x) for x in ver_dic["VERSION"])
 release = ver_dic["VERSION_TEXT"]
 
-extensions += [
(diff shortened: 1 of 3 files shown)


doc/conf.py
```diff
@@ -1,59 +1,23 @@
...
-# ones.
-extensions = [
-    "sphinx.ext.autodoc",
-    "sphinx.ext.doctest",
-    "sphinx.ext.intersphinx",
-    "sphinx.ext.mathjax",
-    "sphinx.ext.graphviz",
-    "sphinx_copybutton",
+extensions += [
+    "sphinx_autodoc_typehints",
```
alexfikl, line 18: Is this required? I was under the impression the typehints functionality is in core sphinx these days?

matthiasdiener: Yeah, it seems to work fine without the extension. Removed.

inducer:
:+1: Thanks!

PR #83, opened 2021-08-26
Add non-recursive reductions

majosm:
Used by #71. Depends on #81 and #82.


arraycontext/container/traversal.py
```diff
@@ -294,6 +296,48 @@ def rec(keys: Tuple[Union[str, int], ...],
...
 # {{{ array container reductions
 
+def map_reduce_array_container(
+        reduce_func: Callable[[Iterable[Any]], Any],
+        map_func: Callable[[Any], Any],
+        ary: ArrayOrContainerT) -> Any:
+    """Perform a map-reduce over array containers.
+
+    :param reduce_func: callable used to reduce over the components of *ary*
+        if *ary* is an :class:`~arraycontext.ArrayContainer`.
```
alexfikl, line 306:
```suggestion
        if *ary* is an :class:`~arraycontext.ArrayContainer`. The callable
        should be associative, as for :func:`rec_map_reduce_array_container`.
```
?

alexfikl (approved):
One small nitpick in the docs, otherwise this looks good to me! :+1:

arraycontext/container/traversal.py
majosm: Hmm... Does it actually need to be associative in this case? I was thinking it didn't, since there isn't any recursion going on. (Not sure if we want to require it anyway, just for consistency with the other functions? 🤷‍♂️)

alexfikl: It feels like it's still worth mentioning something about it. I guess here it would matter more that the operation doing the reduction is associative (not `reduce_func` itself), so that changing the order of the fields of the container (not out of the ordinary for dataclass-based array containers?) will give the same result.

\+ consistency! Always a fan :tada:

inducer:
LGTM, thanks!

Issue #28, opened 2021-06-11
rec_map_array_container: ValueError on trying to traverse a dataclass_array_container with a leaf array


kaushikcfd:
On running:
```python
@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True)
@dataclass_array_container
@dataclass(frozen=True)
class State:
    u: np.ndarray

mystate = State(np.zeros(10))
rec_map_array_container(lambda x: x, mystate)
```

I get an error saying: `ValueError: only object arrays are supported, given dtype 'float64'`. Should dataclass_array_container not contain primitive-dtyped numpy arrays?

I was expecting to get a shallow copy of `mystate` after the traversal.

inducer:
I agree with you; I think this is just being mishandled. The key thing is that while object arrays are array containers, plain numpy arrays are not. So IMO the code should just funnel "normal" numpy arrays down the same path of "not a container" instead of complaining. PR welcome.

PR #90, opened 2021-09-21
Fix map_array_container on non-ArrayContainers

alexfikl:
`map_array_container` was using `try ... except` to check if an object is serializable, but it was only catching `TypeError`, not `NotImplementedError` thrown by `serialize_container`. That basically made it try to deserialize floats and other non-array container objects.

The `try ... except` was introduced in #31, seemingly for some performance reasons (?), so I went ahead and used the same pattern in several other places.

@thomasgibson as this is (now, vaguely) related to inducer/grudge#154.

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -180,7 +180,7 @@ def map_array_container(
     """
     try:
         iterable = serialize_container(ary)
-    except NotImplementedError:
+    except TypeError:
         return f(ary)
     else:
         return deserialize_container(ary, [
@@ -265,7 +265,7 @@ def keyed_map_array_container(f: Callable[[Any, Any], Any],
     """
     try:
         iterable = serialize_container(ary)
-    except NotImplementedError:
+    except TypeError:
         raise ValueError(
                 f"Non-array container type has no key: {type(ary).__name__}")
     else:
@@ -287,7 +287,7 @@ def rec_keyed_map_array_container(f: Callable[[Tuple[Any, ...], Any], Any],
             _ary: ArrayOrContainerT) -> ArrayOrContainerT:
         try:
             iterable = serialize_container(_ary)
-        except NotImplementedError:
+        except TypeError:
             return f(keys, _ary)
         else:
             return deserialize_container(_ary, [
@@ -316,7 +316,7 @@ def map_reduce_array_container(
     """
     try:
         iterable = serialize_container(ary)
-    except NotImplementedError:
+    except TypeError:
         return map_func(ary)
     else:
         return reduce_func([
@@ -391,7 +391,7 @@ def rec_map_reduce_array_container(
     def rec(_ary: ArrayOrContainerT) -> ArrayOrContainerT:
         try:
             iterable = serialize_container(_ary)
-        except NotImplementedError:
+        except TypeError:
             return map_func(_ary)
         else:
             return reduce_func([
@@ -483,7 +483,7 @@ def thaw(ary: ArrayOrContainerT, actx: ArrayContext) -> ArrayOrContainerT:
     """
     try:
         iterable = serialize_container(ary)
-    except NotImplementedError:
+    except TypeError:
         return actx.thaw(ary)
     else:
         return deserialize_container(ary, [
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -179,12 +179,13 @@ def map_array_container(
         or an instance of a base array type.
     """
     try:
-        ser_ctr = serialize_container(ary)
-    except TypeError:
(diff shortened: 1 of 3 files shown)


arraycontext/container/traversal.py
```diff
@@ -179,12 +179,13 @@ def map_array_container(
         or an instance of a base array type.
     """
     try:
-        ser_ctr = serialize_container(ary)
-    except TypeError:
+        iterable = serialize_container(ary)
+    except NotImplementedError:
```
alexfikl, line 183: Is the `TypeError` needed here?

alexfikl: The object array serialization can raise `ValueError` for non-object arrays. Is it worth catching that here?

thomasgibson: So a `ValueError` to me indicates specifically that whatever was passed as `ary` is simply not acceptable. I guess one thing to ask here is what is the formal definition of an "array container". A `NotImplementedError` to me indicates that whatever I passed is an "array container", but is just missing an implementation. So I think if something like a float (I guess a scalar element?) comes in, it should trigger a `ValueError`; floats are not "array containers"

Any thoughts? This is just my mental picture of all this.

arraycontext/container/traversal.py
```diff
@@ -262,12 +263,15 @@ def keyed_map_array_container(f: Callable[[Any, Any], Any],
...
     """
-    if is_array_container(ary):
-        return deserialize_container(ary, [
-                (key, f(key, subary)) for key, subary in serialize_container(ary)
-                ])
+    try:
+        iterable = serialize_container(ary)
+    except NotImplementedError:
+        raise ValueError(
+                f"Non-array container type has no key: {type(ary).__name__}")
```
thomasgibson, line 270: I think we may want to distinguish between `ValueError` and `NotImplementedError`. Also, I'm not sure if we necessarily need to catch exceptions if we do proper error handling further down

arraycontext/container/traversal.py
alexfikl: Currently the `serialize_container` / `deserialize_container` act more like methods on an abstract base class and `is_array_container` just checks if they're implemented for the given type (using [singledispatch.dispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch)).

The default implementation just raises `NotImplementedError` to say "this type does not implement the ABC" and I'm not sure raising `ValueError` makes more sense here. I'm also becoming less convinced that the `ndarray` serialization should raise `NotImplementedError` for other dtypes, since that would cause a mismatch with what `is_array_container` says..

Probably an issue for another day, i.e. #28.

inducer: I'm leaning towards sticking with `TypeError`, actually.

* Why not `ValueError`? That says to me: "the type you're passing is OK, but the specific value/instances of the type you chose is terrible."
* Why not `NotImplementedError`? Agree with @thomasgibson here that that says "eh, we haven't gotten around to it yet." Which may have made sense in the early days of array containers when there were plenty of things that probably should be containers that weren't, but that's no longer today's reality.

> I'm also becoming less convinced that the ndarray serialization should raise `NotImplementedError` for other dtypes, since that would cause a mismatch with what `is_array_container` says.

I think `is_array_container` is just wrong in that case. I'm pretty convinced that plain-old numpy arrays aren't containers. I'd be more inclined to redefine `is_array_container` as "try to serialize it, and it's not a container if that fails". We're on that road with this PR anyway.

inducer: I.e. what I mean here is to change the default implementation of `{de,}serialize_container` to raise `TypeError`.

alexfikl: Done in baaab74.

Agreed that `is_array_container` not quite doing its job and we should just let `serialize_container` figure out if it can do it or not.

arraycontext/container/traversal.py
alexfikl: Not quite sure I follow what the suggestion is. Can you elaborate a bit? This code should be the same as before, it just uses `try .. except` instead of `is_array_container`.

inducer:
LGTM! I'll go ahead and merge, since I don't think there are any total dealbreakers in here. If there are lingering concerns (@thomasgibson?), please bump them to an issue.

arraycontext/container/traversal.py
thomasgibson: @alexfikl I think the discussion in this thread https://github.com/inducer/arraycontext/pull/90#discussion_r714292397 settles this!

alexfikl: Got it! :grin: 

Thanks a lot for taking a look and sparking a discussion! Always appreciated!

PR #95, opened 2021-09-28
Deprecate with_container_arithmetic's bcast_numpy_array arg

kaushikcfd:
```
Passing both 'bcast_numpy_array' and '_bcast_actx_array_types' was
ill-defined. For example, in the case of an ArrayContext whose thawed
array type is np.ndarray the specification would contradict between
broadcasting the argument numpy_array to return an object array *OR*
peforming the operation with every leaf array.

Consider the example below,

(
- 'Foo: ArrayContainer' whose arithmetic routines are
   generated by `with_container_arithmetic(bcast_numpy=True,
   _bcast_actx_array_types=True)`
- 'actx: ArrayContextT' for whom `np.ndarray` is a valid thawed array
   type.
)

Foo(DOFArray(actx, [38*actx.ones(3, np.float64)])) + np.array([3, 4, 5])
could be either of:
- array([Foo(DOFArray([array([41, 41, 41])])),
         Foo(DOFArray([array([42, 42, 42])])),
         Foo(DOFArray([array([43, 43, 43])]))]), OR,
- Foo(DOFArray(actx, array([41, 42, 43])))
```

Draft because
- [ ] https://github.com/ecisneros8/pyrometheus/pull/35
- [ ] https://github.com/illinois-ceesd/mirgecom/pull/514

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -206,6 +206,15 @@ def with_container_arithmetic(
     if rel_comparison is None:
         raise TypeError("rel_comparison must be specified")
 
+    if bcast_numpy_array:
+        from warnings import warn
+        warn("'bcast_numpy_array=True' is deprecated and will be unsupported"
+             " from December 2021", DeprecationWarning, stacklevel=2)
+
+        if _bcast_actx_array_type:
+            raise ValueError("'bcast_numpy_array' and '_bcast_actx_array_type'"
+                             " cannot be both set.")
+
     if rel_comparison and eq_comparison is None:
         eq_comparison = True
 
@@ -216,7 +225,7 @@ def with_container_arithmetic(
         raise TypeError("bcast_obj_array must be set if bcast_numpy_array is")
 
     if _bcast_actx_array_type is None:
-        if _cls_has_array_context_attr:
+        if _cls_has_array_context_attr and (not bcast_numpy_array):
             _bcast_actx_array_type = bcast_number
         else:
             _bcast_actx_array_type = False


majosm:
Not sure if I like the change in illinois-ceesd/mirgecom#514... requiring the application programmer to distinguish between numpy arrays of dtype float and object arrays containing floats seems less than ideal.

Would it be enough to just forbid setting both `bcast_numpy_array` and `_bcast_actx_array_type` to `True`? Maybe `_bcast_actx_array_type` shouldn't be `True` by default?

kaushikcfd:
@majosm: Yep, the interface won't be as brief as before, but I can see folks preferring Numpy over PyOpenCL in MIRGE-Com (for development purposes), as PyOpenCL arrays aren't quite fully implemented when it comes to features like broadcasting, non-contiguous copies, etc. (@thomasgibson took this route while prototyping some of his stuff).

And, if we intend to support a Numpy-based array context, casting them into object arrays before using them for arithmetic is a deal-breaker for its implementation.

arraycontext/container/arithmetic.py
```diff
@@ -206,6 +206,15 @@ def _deserialize_init_arrays_code(cls, tmpl_instance_name, args):
     if rel_comparison is None:
         raise TypeError("rel_comparison must be specified")
 
+    if bcast_numpy_array:
+        from warnings import warn
+        warn("'bcast_numpy_array=True' is deprecated and will be unsupported"
+             " from November, 2021", DeprecationWarning, stacklevel=2)
```
inducer, line 212:
```suggestion
             " from December 2021", DeprecationWarning, stacklevel=2)
```

inducer:
I think an array context that just wraps `numpy` is well worth having. (I can already think of quite a few use cases myself, and @thomasgibson's case demonstrates that we'd want to allow that even in mirgecom.)

As far as I can tell, it's not straightforward from looking at a numpy array in `bcast_numpy_array` processing whether the array is intended to be a container (and thus broadcast) or not. Forcing those arrays to be object arrays is sensible IMO, even if it makes the user code less slick-looking. Again IMO, we erred too much on the side of magic here and scored a bit of an own goal in terms of backward compatibility.

If there's an elegant way that gets us numpy array contexts and avoids the compat break, I'd be curious to hear about it. But as long as it an either-or between plain-numpy array contexts and slicker-looking broadcasting, I'll take the former, I think.

arraycontext/container/arithmetic.py
kaushikcfd: Thanks, see  7da8e1f.

kaushikcfd:
> Is there a use case for _bcast_actx_array_type=True for things like DOFArray and ConservedVars? I just tried disabling it, and it didn't seem to break anything.

Yep, if we use PytatoArrayContext for the project https://github.com/illinois-ceesd/drivers_y1-nozzle, we do see failures  if  `not _bcast_actx_array_type`. As the compiled operator makes "time" (a scalar in PyOpenCLArrayContext) as a placeholder array, and, there are [locations](https://github.com/illinois-ceesd/drivers_y1-nozzle/blob/1e4d3991e2901e6eeed30e943f090f0616d26634/nozzle.py#L411-L416) in the nozzle-driver where the "time" variable must be broadcasted to all the arrays in a container.

kaushikcfd:
About `_bcast_actx_array_type`: there was also some discussion in https://github.com/inducer/arraycontext/issues/49.

PR #96, opened 2021-10-05
Allow to manually exclude fields in dataclass_array_container

alexfikl:
This isn't as nice as it could be due to backwards compatibility, but it seems to work.

My use case is trying to time step a variable that has some additional fields (updated at the same time from the variable) that would be nice to drag along, without involving them in the arithmetic.


alexfikl:
@inducer Tried to use that `Annotated` type to explicitly exclude fields from the dataclass and it seems to be working quite nicely. Don't have a Python 3.6 environment to check that it works there too, but it should if `typing_extensions` works as advertised.

PR #107, opened 2021-10-18
Deprecate is_array_container

### LINKED ISSUE #106: Deprecate `is_array_container`
I think it's confusing because it only looks at the type. That means, for example for non-object numpy arrays, it can give the wrong answer.

cc @majosm @alexfikl

inducer:
Closes #106.

### DIFF
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -22,7 +22,7 @@
 
 Serialization/deserialization
 -----------------------------
-.. autofunction:: is_array_container
+.. autofunction:: is_array_container_type
 .. autofunction:: serialize_container
 .. autofunction:: deserialize_container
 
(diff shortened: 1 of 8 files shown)


arraycontext/container/__init__.py
```diff
@@ -168,6 +173,13 @@ def is_array_container(ary: Any) -> bool:
     :returns: *True* if the instance *ary* has a registered implementation of
         :func:`serialize_container`.
     """
+
+    from warnings import warn
+    warn("is_array_container is deprecated and will be removed in 2022. "
+            "If you must know precisely whether something is an array container, "
+            "try serializing it and catch TypeError. For a cheaper option, see "
```
alexfikl, line 180: If I recall correctly, the `ndarray` serialization didn't raise a `TypeError` when it found a non-object array one (maybe `ValueError`?). Should probably change that to be consistent.

Possibly out of scope, but how about introducing some `ContainerSerializationError` / `NotArrayContainerError` / `somethingError` to catch these?

alexfikl (approved):
Just one small issue, otherwise this looks good to me!

arraycontext/container/__init__.py
inducer: Good point! 557f9dd

Not opposed to a separate error type, but IMO out of scope for now.

PR #108, opened 2021-10-19
FromActxCompile tag

kaushikcfd:


kaushikcfd:
My current approach is to have have a separate transform strategy for the operator kernels than the leaner frozen kernels. This tag is helpful to branch on that. (As taken in https://github.com/inducer/grudge/blob/dc7cde20d6481917103f0163a69dc0c97e12d757/examples/wave/wave-opt-3d.py#L497)

arraycontext/impl/pytato/compile.py
```diff
@@ -40,13 +40,25 @@
...
 import itertools
+from pytools.tag import Tag
 
 from pytools import ProcessLogger
 
 import logging
 logger = logging.getLogger(__name__)
 
 
+class FromActxCompile(Tag):
```
inducer, line 51: Sorry, spotted another thing: This should be hooked into the docs.

alexfikl: `FromArrayContextCompile`? :grin:

PR #110, opened 2021-10-19
Implements TaggableCLArray

kaushikcfd:
Converting to draft because #111, will soon make array axes taggable as well, which the `TaggableCLArray` should record as well.

- [ ] Must go after #111


kaushikcfd:
Closing because got merged into #111.

PR #111, opened 2021-10-20
Defines TaggableCLArray

kaushikcfd:
- [x] Must go after https://github.com/inducer/pytato/pull/170
- [x] Point the correct branches in `requirements.txt`.

### DIFF
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -1,6 +1,7 @@
 """
 .. currentmodule:: arraycontext
 .. autoclass:: PyOpenCLArrayContext
+.. automodule:: arraycontext.impl.pyopencl.taggable_cl_array
 """
 
 __copyright__ = """
@@ -35,6 +36,7 @@ import numpy as np
 from pytools.tag import Tag
 
 from arraycontext.context import ArrayContext, _ScalarLike
+from arraycontext.container.traversal import rec_map_array_container
 
 
 if TYPE_CHECKING:
@@ -65,6 +67,8 @@ class PyOpenCLArrayContext(ArrayContext):
         of arrays are created (e.g. as results of computation), the associated cost
         may become significant. Using e.g. :class:`pyopencl.tools.MemoryPool`
         as the allocator can help avoid this cost.
+
+    .. automethod:: transform_loopy_program
     """
 
     def __init__(self,
@@ -109,7 +113,7 @@ class PyOpenCLArrayContext(ArrayContext):
                     DeprecationWarning, stacklevel=2)
 
         import pyopencl as cl
-        import pyopencl.array as cla
+        import pyopencl.array as cl_array
 
         super().__init__()
         self.context = queue.context
@@ -138,7 +142,9 @@ class PyOpenCLArrayContext(ArrayContext):
         self._loopy_transform_cache: \
                 Dict["lp.TranslationUnit", "lp.TranslationUnit"] = {}
 
-        self.array_types = (cla.Array,)
+        # TODO: Ideally this should only be `(TaggableCLArray,)`, but
+        # that would break the logic in the downstream users.
+        self.array_types = (cl_array.Array,)
 
     def _get_fake_numpy_namespace(self):
         from arraycontext.impl.pyopencl.fake_numpy import PyOpenCLFakeNumpyNamespace
@@ -147,18 +153,27 @@ class PyOpenCLArrayContext(ArrayContext):
     # {{{ ArrayContext interface
 
     def empty(self, shape, dtype):
-        import pyopencl.array as cl_array
-        return cl_array.empty(self.queue, shape=shape, dtype=dtype,
-                allocator=self.allocator)
+        from arraycontext.impl.pyopencl.taggable_cl_array import TaggableCLArray
+
+        return TaggableCLArray(self.queue, shape=shape, dtype=dtype,
+                               allocator=self.allocator)
 
     def zeros(self, shape, dtype):
         import pyopencl.array as cl_array
-        return cl_array.zeros(self.queue, shape=shape, dtype=dtype,
-                allocator=self.allocator)
+        from arraycontext.impl.pyopencl.taggable_cl_array import to_tagged_cl_array
+        return to_tagged_cl_array(cl_array.zeros(self.queue, shape=shape,
+                                                 dtype=dtype,
+                                                 allocator=self.allocator),
+                                  axes=None, tags=frozenset())
 
     def from_numpy(self, array: Union[np.ndarray, _ScalarLike]):
         import pyopencl.array as cl_array
-        return cl_array.to_device(self.queue, array, allocator=self.allocator)
+        from arraycontext.impl.pyopencl.taggable_cl_array import to_tagged_cl_array
+        return to_tagged_cl_array(cl_array
+                                  .to_device(self.queue,
+                                             array,
+                                             allocator=self.allocator),
+                                  axes=None, tags=frozenset())
 
     def to_numpy(self, array):
         if np.isscalar(array):
@@ -186,14 +201,33 @@ class PyOpenCLArrayContext(ArrayContext):
             if len(wait_event_queue) > self._wait_event_queue_length:
                 wait_event_queue.pop(0).wait()
 
-        return result
+        from arraycontext.impl.pyopencl.taggable_cl_array import to_tagged_cl_array
+        # FIXME: Inherit loopy tags for these arrays
+        return {name: to_tagged_cl_array(ary, axes=None, tags=frozenset())
+                for name, ary in result.items()}
 
     def freeze(self, array):
         array.finish()
         return array.with_queue(None)
 
     def thaw(self, array):
-        return array.with_queue(self.queue)
+        from arraycontext.impl.pyopencl.taggable_cl_array import (TaggableCLArray,
+                                                                  to_tagged_cl_array)
+        import pyopencl.array as cl_array
+
+        if isinstance(array, TaggableCLArray):
+            return array.with_queue(self.queue)
+        elif isinstance(array, cl_array.Array):
+            from warnings import warn
+            warn("Invoking PyOpenCLArrayContext.thaw with pyopencl.Array"
+                 " will be unsupported in 2023. Use `to_tagged_cl_array`"
+                 " to convert instances of pyopencl.Array to TaggableCLArray.",
+                 DeprecationWarning, stacklevel=2)
+            return (to_tagged_cl_array(array, axes=None, tags=frozenset())
+                    .with_queue(self.queue))
+        else:
+            raise ValueError("array should be a cl.array.Array,"
+                             f" got '{type(array)}'")
 
     # }}}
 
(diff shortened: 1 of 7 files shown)


arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,112 @@
...
+
+.. autofunction:: to_tagged_cl_array
+"""
+
+import pyopencl.array as cla
+from typing import FrozenSet, Union, Sequence, Optional, Tuple
+from pytools.tag import Taggable, Tag, tag_dataclass
+
+
+@tag_dataclass
```
inducer, line 13: I'm not sure this is itself a tag, so I don't think the `@tag_dataclass` belongs.

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,112 @@
...
+                         order=order, allocator=allocator,
+                         data=data, offset=offset,
+                         strides=strides, events=events,
+                         _flags=_flags, _fast=_fast,
+                         _size=_size, _context=_context,
+                         _queue=_queue)
+
+        self.tags = tags
+        axes = axes if axes is not None else tuple(Axis(frozenset())
+                                                   for _ in self.shape)
```
inducer, line 44: I think it might be better to expect this to be in place, cf. `to_tagged_cl_array`, in  order to keep this constructor fast. It was already one of the hottest code paths in eager eval.

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -35,11 +36,13 @@
...
+from arraycontext.container.traversal import rec_map_array_container
 
 
 if TYPE_CHECKING:
     import pyopencl
     import loopy as lp
+    from arraycontext.impl.pyopencl.taggable_cl_array import TaggableCLArray
 
 
 # {{{ PyOpenCLArrayContext
```
inducer, line 48: Maybe also augment `PyOpenCLArrayContext.array_types`?

arraycontext/impl/pytato/__init__.py
```diff
@@ -112,13 +114,13 @@ def to_numpy(self, array):
...
-        import pyopencl.array as cla
         from pytato.loopy import call_loopy
+        from arraycontext.impl.pyopencl.taggable_cl_array import TaggableCLArray
 
         entrypoint = program.default_entrypoint.name
 
         # thaw frozen arrays
-        kwargs = {kw: (self.thaw(arg) if isinstance(arg, cla.Array) else arg)
+        kwargs = {kw: (self.thaw(arg) if isinstance(arg, TaggableCLArray) else arg)
                   for kw, arg in kwargs.items()}
```
inducer, line 124: I think it'd be wise to allow tagged and untagged arrays, here and below. (Or, at least, complain about untagged ones.)

inducer (commented):
Thanks! Some thoughts from a first scroll below.

inducer:

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,112 @@
...
+    tags: FrozenSet[Tag]
+
+    def copy(self, **kwargs):
+        from dataclasses import replace
+        return replace(self, **kwargs)
+
+
+class TaggableCLArray(cla.Array, Taggable):
+    """
+    A :class:`pyopencl.array.Array` with additional attributes record
```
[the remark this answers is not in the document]
kaushikcfd: Done, please let me know if still not clear.

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Oops, yep. Amended.

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Sorry, I don't think that's possible. After doing any pyopencl operation we instantiate a new TaggableCLArray which would mean that we would carry along a `None` for the axes. Tried to keep the costs low by using `pytools.memoize`.

arraycontext/impl/pytato/__init__.py
kaushikcfd: I don't see a good reason to allow untagged arrays. Just made sure that all the arrays had the correct types.

kaushikcfd:
Tests should pass once https://github.com/inducer/pyopencl/pull/531 (or something equivalent) lands.

alexfikl:
Any chance this works with `pytential` too? That has a very "interesting" combination of using array contexts and direct calls to `pyopencl`, which is easy to break.

Maybe worth adding to the CI?

kaushikcfd:
> and direct calls to pyopencl

:nauseated_face: 

Thanks, added a CI here: https://github.com/kaushikcfd/pytential/pull/1. Will see if anything "interesting" pops up.

kaushikcfd:
@alexfikl: Thanks for catching that. 

> ===== 136 failed, 65 passed, 1 skipped, 385 warnings in 828.91s (0:13:48) ======

There are quite a few failures in `pytential`.


Probably not a smart thing to require this type. So there are a couple of options I can do here:
1. Fix pytential to not do this. (Unsure how involved this would be)
2. Just keep this limited to PytatoPyOpenCLArrayContext.

kaushikcfd:
> I imagine it would be quite a bit of work to get pytential to work. Besides calling pyopencl.array directly, it also calls loopy directly (without actx.call_loopy) and a few algorithms from pyopencl (ElementwiseKernel?).. so, yeah, a bit of work!

Ah, thanks! Yep, I'll just go with not requiring `TaggableCLArray` in PyOpenCLArrayContext.

kaushikcfd:
> not requiring TaggableCLArray in PyOpenCLArrayContext.

Yep, doing that seems to fix the CIs in potential as well! (Thanks!)

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,127 @@
...
+            axes = self.axes if axes is None else axes
+            return self.__class__(None, self.shape, self.dtype,
+                                  allocator=self.allocator,
+                                  strides=self.strides, data=self.base_data,
+                                  offset=self.offset, events=self.events,
+                                  _fast=True, _context=self.context,
+                                  _queue=self.queue, _size=self.size,
+                                  tags=tags, axes=axes)
+        else:
+            new_with_queue = super().copy(queue=queue)
```
inducer, line 80: Why this redundant `copy`? Could you add a comment to explain?

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,127 @@
...
+            return self.__class__(None, new_with_queue.shape,
```
inducer, line 81: Is the code duplication of the two constructor calls here avoidable? (Maybe by setting a `base_instance` variable?)

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+@memoize
+def _construct_untagged_axes(ndim: int) -> Tuple[Axis, ...]:
+    return tuple(Axis(frozenset()) for _ in range(ndim))
+
+
+class TaggableCLArray(cla.Array, Taggable):
+    """
+    A :class:`pyopencl.array.Array` with additional attributes record
```
inducer, line 34:
```suggestion
    A :class:`pyopencl.array.Array` with additional metadata. This is used by
    :class:`PytatoPyOpenCLArrayContext` to preserve tags for data while frozen,
    and also in a similar capacity by :class:`PyOpenCLArrayContext`.
```

arraycontext/impl/pytato/__init__.py
```diff
@@ -112,39 +114,65 @@ def to_numpy(self, array):
...
         import pytato as pt
         import pyopencl.array as cla
         import loopy as lp
+        from arraycontext.impl.pytato.utils import (_normalize_pt_expr,
+                                                    get_cl_axes_from_pt_axes)
+        from arraycontext.impl.pyopencl.taggable_cl_array import (to_tagged_cl_array,
+                                                                  TaggableCLArray)
 
-        if isinstance(array, cla.Array):
+        if isinstance(array, TaggableCLArray):
```
inducer, line 152: Why does `cla.Array` now fall through to the bottom of this?

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -0,0 +1,127 @@
...
+                           ary.dtype,
+                           allocator=ary.allocator,
+                           strides=ary.strides,
+                           data=ary.base_data,
+                           offset=ary.offset,
+                           events=ary.events, _fast=True,
+                           _context=ary.context,
+                           _queue=ary.queue, _size=ary.size,
+                           axes=axes,
+                           tags=tags)
```
inducer, line 127: I don't love the idea of hardcoding so much of the internal structure of `cla.Array` here. Is there a copy method in `Array` that we could use instead?

arraycontext/impl/pytato/__init__.py
```diff
@@ -165,17 +199,31 @@ def freeze(self, array):
...
-        return pt.make_data_wrapper(array.with_queue(self.queue))
+        from .utils import get_pt_axes_from_cl_axes
+        from arraycontext.impl.pyopencl.taggable_cl_array import (TaggableCLArray,
+                                                                  to_tagged_cl_array)
+        import pyopencl.array as cl_array
+
+        if isinstance(array, TaggableCLArray):
+            pass
+        elif isinstance(array, cl_array.Array):
+            array = to_tagged_cl_array(array, axes=None, tags=frozenset())
```
inducer, line 218: Maybe deprecate this usage?

inducer (commented):
Thanks! Good to go after these fixes.

arraycontext/impl/pytato/__init__.py
kaushikcfd: Done, thanks!

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Not quite, see [cla.Array.copy](https://github.com/inducer/pyopencl/blob/fb13a7dfe4c6c07b0661fb554663bc403c28b97c/pyopencl/array.py#L828). But tried to avoid the duplication by invoking `TaggableCLArray.copy`. So there's only one source location where we rely on the structure on `cl.array.Array.__init__`'s signature.

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Done, thanks!

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Indeed, using ` base_instance` was helpful.

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: Done.

arraycontext/impl/pyopencl/taggable_cl_array.py
inducer: OK, I'll take this. At a later point, we may decide that we can save some function calls by inlining, but let's wait with that.

arraycontext/impl/pytato/__init__.py
```diff
@@ -207,23 +263,22 @@ def transform_dag(self, dag: "pytato.DictOfNamedArrays"
...
     def einsum(self, spec, *args, arg_names=None, tagged=()):
-        import pyopencl.array as cla
+        from arraycontext.impl.pyopencl.taggable_cl_array import TaggableCLArray
         import pytato as pt
         if arg_names is None:
             arg_names = (None,) * len(args)
 
         def preprocess_arg(name, arg):
-            if isinstance(arg, cla.Array):
+            if isinstance(arg, TaggableCLArray):
```
inducer, line 281: If you'd like to disallow passing in `cla.Array` here, raise an appropriate error instead of falling through.

arraycontext/impl/pytato/compile.py
```diff
@@ -399,7 +422,12 @@ def __call__(self, arg_id_to_arg) -> ArrayContainer:
         evt.wait()
 
         def to_output_template(keys, _):
-            return self.actx.thaw(out_dict[self.output_id_to_name_in_program[keys]])
+            name_in_program = self.output_id_to_name_in_program[keys]
+            return self.actx.thaw(to_tagged_cl_array(
+                out_dict[name_in_program],
+                axes=get_cl_axes_from_pt_axes(
+                    self.name_in_program_to_axes[name_in_program]),
```
inducer, line 429: Could store the CL axes instead, to avoid this conversion cost.

arraycontext/impl/pytato/__init__.py
kaushikcfd: Yep, added a warning here too.

arraycontext/impl/pytato/compile.py
kaushikcfd: Not without some type-incoherencies, recorded it over here: https://github.com/inducer/pytato/issues/261.

inducer: Makes sense.

inducer:
LGTM, thanks!

PR #112, opened 2021-10-20
Convert some is_array_container_type checks

### LINKED ISSUE #28: rec_map_array_container: ValueError on trying to traverse a dataclass_array_container with a leaf array

alexfikl:
Converts most of the `is_array_container_type` checks to use `try: serialize except: ...` instead, so that the various traversal functions do the same thing.

Fixes #28


arraycontext/container/__init__.py
```diff
@@ -115,6 +117,10 @@ class ArrayContainer:
     """
 
 
+class ArrayContainerTypeError(TypeError):
```
alexfikl, line 120: @inducer Added this in 3499359. Not sold on the name, but now it should be easy to `sed` for it.

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -239,32 +240,33 @@ def all(self, a):
...
     def array_equal(self, a, b):
-        def as_device_scalar(bool_value):
-            import numpy as np
-            return self._array_context.from_numpy(
-                np.array(int(bool_value), dtype=np.int8))
+        actx = self._array_context
+        queue = actx.queue
+
+        # NOTE: pyopencl doesn't like `bool` much, so use `int8` instead
+        false = actx.from_numpy(np.int8(False))
```
alexfikl, line 247: After inducer/pyopencl#518 and #113, I'm not sure this needs to be `from_numpy`. Thoughts?

alexfikl: Sounds good!

I'm not too bothered either way, since, like `norm`, the rule would be "if we don't do any device computation, we return a host thing", which seems reasonable enough.

Issue #101, opened 2021-10-12
Handling temperature initial guesses


inducer:
In full-on combustion mirgecom, temperature is found via Newton iteration in https://github.com/ecisneros8/pyrometheus. This is expensive, and it's made substantially cheaper/more reliable by the availability of a starting guess (typically, the last temperature). In eager eval, this is easy: the last value can be sent "along for the ride" in the `ConservedVars` and looked at when needed. There are two aspects here:

* So it sits in an array container, but it may not participate in arithmetic. (Realize that it's possible that no meaningful "temperature increments" are being computed, so the array container parts related to temperature initial guesses returned for the ODE RHS are likely `None`. The container arithmetic should be able to take that. Conceivably we could compute those increments, and the arithmetic *might* start making sense... but I'm not sure we want to.) So this is the first question: Do we teach `with_container_arithmetic` about what's needed? How? (Include `Optional` in the field type spec? More arguments like `allow_none=["field1", "field2"]`?)
* The other aspect is that this temperature initial guess needs to be "placeholderized" and propagated into `actx.compile`d code. As long as it's a somewhat normal part of an array container, that happens automatically. (but @mtcam: the "just copy it from the template" mode we discussed likely won't work with lazy, because it won't get "placeholderized" upon entry to a `compile`d function.)

cc @mtcam @kaushikcfd @alexfikl

PR #116, opened 2021-10-26
Fixes for dataclass_array_container

alexfikl:
There's a couple of random fixes to the dataclass helper in here:

1. Add an assert in `is_array_container_type` that `isinstance(cls, type)`. If a dataclass has an `Optional` or `Union` field (which aren't classes), `is_array_container_type` would raise some `AttributeError` anyway, but with a more cryptic message.

2. Added some checks in `dataclass_array_container` to give better error messages when it encounters unsuported field types: with `init=False`, string annotations or classes wrapped in `Optional` (or equivalent).

### DIFF
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -173,9 +173,7 @@ def is_array_container_type(cls: type) -> bool:
         function will say that :class:`numpy.ndarray` is an array container
         type, only object arrays *actually are* array containers.
     """
-    assert isinstance(cls, type), \
-            f"must pass a type, not an instance: '{cls!r}'"
-    assert hasattr(cls, "__mro__"), "'cls' has no attribute '__mro__': "
+    assert isinstance(cls, type), f"must pass a {type!r}, not a '{cls!r}'"
 
     return (
             cls is ArrayContainer
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -173,6 +173,10 @@ def is_array_container_type(cls: type) -> bool:
         function will say that :class:`numpy.ndarray` is an array container
         type, only object arrays *actually are* array containers.
     """
+    assert isinstance(cls, type), \
+            f"must pass a type, not an instance: '{cls!r}'"
+    assert hasattr(cls, "__mro__"), "'cls' has no attribute '__mro__': "
+
     return (
             cls is ArrayContainer
             or (serialize_container.dispatch(cls)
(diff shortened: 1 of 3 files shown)


arraycontext/container/__init__.py
```diff
@@ -173,6 +173,10 @@ def is_array_container_type(cls: type) -> bool:
         function will say that :class:`numpy.ndarray` is an array container
         type, only object arrays *actually are* array containers.
     """
+    assert isinstance(cls, type), \
+            f"must pass a type, not an instance: '{cls!r}'"
+    assert hasattr(cls, "__mro__"), "'cls' has no attribute '__mro__': "
```
alexfikl, line 178: Actually, is this `__mro__` check needed? Everything that is an instance of `type` should have an `__mro__` defined, right?

arraycontext/container/dataclass.py
```diff
@@ -48,11 +48,15 @@ def dataclass_array_container(cls: type) -> type:
     """
     from dataclasses import is_dataclass
     assert is_dataclass(cls)
+    assert all(not isinstance(f.type, str) for f in fields(cls)), \
+            "dataclass cannot have string-based type annotations"
 
     array_fields = [
-            f for f in fields(cls) if is_array_container_type(f.type)]
+            f for f in fields(cls)
+            if f.init and is_array_container_type(f.type)]
```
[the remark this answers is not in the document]
alexfikl: @inducer For the moment, reworked all of this to a bunch of stricter error checks and better error messages in e1d0dc6. Let me know what you think!

Fixing #101 seems to be more involved, so leaving that for its own PR.

arraycontext/container/__init__.py
inducer: I don't think it's needed.

Just to check, `isinstance(..., type)` does seem to successfully rule out the `typing` stuff:
```
>>> from typing import Sequence
>>> isinstance(Sequence, type)
False
```

inducer (commented):
Thanks! LGTM with the MRO check removed.

arraycontext/container/__init__.py
alexfikl: Thanks for checking! Removed in 1a28536.

arraycontext/container/__init__.py
```diff
@@ -173,7 +173,12 @@ def is_array_container_type(cls: type) -> bool:
         function will say that :class:`numpy.ndarray` is an array container
         type, only object arrays *actually are* array containers.
     """
-    return (
+    # NOTE: `is_array_container_type` is used in `dataclass_array_container`
+    # where it looks at field types. The fields can contain typing information
+    # like `typing.Optional[MyType]`, which does not have an `__mro__` and messes
+    # up `singledispatch.dispatch`, so we check for that first
```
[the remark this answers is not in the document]
inducer: Those look good to me!

inducer:
Thanks!

PR #121, opened 2021-11-07
Use np.isscalar instead of isinstance(x, Number)

alexfikl:
There's about a dozen places checking for scalars (from a quick grep) and these were the only ones using `isinstance(x, Number)`, so this should make it more consistent.


inducer:
Thx!

PR #122, opened 2021-11-07
Fix scalars case in loopy_implemented_elwise_func

alexfikl:
- [x] inducer/pytato#202

### DIFF
--- a/arraycontext/fake_numpy.py
+++ b/arraycontext/fake_numpy.py
@@ -170,7 +170,8 @@ class BaseFakeNumpyNamespace:
         c_name = self._numpy_to_c_arc_functions.get(name, name)
 
         # limit which functions we try to hand off to loopy
-        if name in self._numpy_math_functions:
+        if (name in self._numpy_math_functions
+                or name in self._c_to_numpy_arc_functions):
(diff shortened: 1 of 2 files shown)


arraycontext/fake_numpy.py
```diff
@@ -160,15 +162,16 @@ def loopy_implemented_elwise_func(*args):
...
                     "The old name will stop working in 2021.",
                     DeprecationWarning, stacklevel=3)
 
         # normalize to C names anyway
         c_name = self._numpy_to_c_arc_functions.get(name, name)
 
         # limit which functions we try to hand off to loopy
-        if name in self._numpy_math_functions:
+        if (name in self._numpy_math_functions
+                or name in self._c_to_numpy_arc_functions):
```
alexfikl, line 174: Not quite sure if this was a bug or intentially skipped?

As far as I can tell, before calling `actx.np.atan` would raise the warning above and then say the `atan` is not implemented.

inducer: No, I think this was just overlooked and never actually used. Thanks for spotting (and testing!) it.

PR #124, opened 2021-11-18
Refactor PytatoActx.compile for usability by distributed

inducer:
@kaushikcfd This shuffles some stuff around and renames some stuff, to allow reuse of various bits in the distributed context. There should be no functional changes, unless we screwed something up. The diff is a bit useless (sorry!), but I hope the intent is still somewhat recognizable.

cc @matthiasdiener


kaushikcfd (approved):
Looks mostly pulling some more code into routines and better naming. LGTM!

PR #126, opened 2021-11-26
PytatoPyOpenCLArrayContext: always invoke MPMS materializer

kaushikcfd:
Suggested by @inducer in https://github.com/inducer/meshmode/pull/254#discussion_r756490536. It seems like a reasonable default that we try to cut down on the number of FlOps in the compiled kernel.


inducer:
Thx!

PR #127, opened 2021-11-29
Distributed execution context

inducer:
- I'm not at all sure whether this should live in `arraycontext`, but I also don't know of a better place. It's clearly related.
- Names?
- Interfaces? So far, this is a weak mockup of what might be needed to unify what grudge needs from eager and lazy

cc @ZwFink @matthiasdiener

### DIFF
--- a/arraycontext/distributed/__init__.py
+++ b/arraycontext/distributed/__init__.py
@@ -1,0 +1,134 @@
+"""
+.. autoclass:: DistributedExecutionContext
+"""
+
+__copyright__ = """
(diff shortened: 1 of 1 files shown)


ZwFink:
Can you provide a little more context for how this ```DistributedExecutionContext``` would be used, and in what ways it would interact with the existing communication code in, e.g., meshmode? Is this intended to hide communication details of lazy execution?

inducer:
> Is this intended to hide communication details of lazy execution?

As discussed in the meeting today, yes, that's exactly the objective.

PR #128, opened 2021-12-13
Allow specifying leaf class in recursive map and map-reduce

majosm:
I think this could help simplify the logic in a couple of places in illinois-ceesd/mirgecom#566 ([example](https://github.com/illinois-ceesd/mirgecom/blob/6e2e4d3e1178de7190ffb17ec21cf4d7732c5137/mirgecom/filter.py#L204-L209)).

cc @thomasgibson

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -256,7 +256,8 @@ def multimap_array_container(f: Callable[..., Any], *args: Any) -> Any:
 
 def rec_map_array_container(
         f: Callable[[Any], Any],
-        ary: ArrayOrContainerT) -> ArrayOrContainerT:
+        ary: ArrayOrContainerT,
+        leaf_class: Optional[type] = None) -> ArrayOrContainerT:
     r"""Applies *f* recursively to an :class:`ArrayContainer`.
 
     For a non-recursive version see :func:`map_array_container`.
@@ -264,18 +265,23 @@ def rec_map_array_container(
     :param ary: a (potentially nested) structure of :class:`ArrayContainer`\ s,
         or an instance of a base array type.
     """
-    return _map_array_container_impl(f, ary, recursive=True)
+    return _map_array_container_impl(f, ary, leaf_cls=leaf_class, recursive=True)
 
 
 def mapped_over_array_containers(
-        f: Callable[[Any], Any]) -> Callable[[ArrayOrContainerT], ArrayOrContainerT]:
+        f: Callable[[Any], Any],
+        leaf_class: Optional[type] = None) -> Callable[
+            [ArrayOrContainerT], ArrayOrContainerT]:
     """Decorator around :func:`rec_map_array_container`."""
-    wrapper = partial(rec_map_array_container, f)
+    wrapper = partial(rec_map_array_container, f, leaf_class=leaf_class)
     update_wrapper(wrapper, f)
     return wrapper
 
 
-def rec_multimap_array_container(f: Callable[..., Any], *args: Any) -> Any:
+def rec_multimap_array_container(
+        f: Callable[..., Any],
+        *args: Any,
+        leaf_class: Optional[type] = None) -> Any:
     r"""Applies *f* recursively to multiple :class:`ArrayContainer`\ s.
 
     For a non-recursive version see :func:`multimap_array_container`.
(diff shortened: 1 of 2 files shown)


arraycontext/container/traversal.py
```diff
@@ -256,43 +256,51 @@ def multimap_array_container(f: Callable[..., Any], *args: Any) -> Any:
...
 def mapped_over_array_containers(
-        f: Callable[[Any], Any]) -> Callable[[ArrayOrContainerT], ArrayOrContainerT]:
+        f: Callable[[Any], Any],
+        leaf_class: Optional[type] = None) -> Callable[
+            [ArrayOrContainerT], ArrayOrContainerT]:
     """Decorator around :func:`rec_map_array_container`."""
-    wrapper = partial(rec_map_array_container, f)
+    wrapper = partial(rec_map_array_container, f, leaf_class=leaf_class)
     update_wrapper(wrapper, f)
     return wrapper
```
alexfikl, line 278: I don't think this is going to work as expected. Is it supposed to be used like
```python
@mapped_over_array_containers(leaf_class=DOFArray)
def my_function(ary):
	...
```
? In that case it should have another wrapper inside (there's an example in [here](https://www.pythontutorial.net/advanced-python/python-decorator-arguments/) for decorators with arguments).

alexfikl (commented):
One little comment, but codewise this looks good to me!

What's the use case? Looking at illinois-ceesd/mirgecom#566, it seems like it would only be used with `DOFArray`s, but there's already a function for that in [meshmode.dof_array](https://github.com/inducer/meshmode/blob/a7a8a060bb3103abcf1c4e509f72d6b324384d71/meshmode/dof_array.py#L359-L378). Would that do the trick?

majosm:
> What's the use case? Looking at [illinois-ceesd/mirgecom#566](https://github.com/illinois-ceesd/mirgecom/pull/566), it seems like it would only be used with `DOFArray`s, but there's already a function for that in [meshmode.dof_array](https://github.com/inducer/meshmode/blob/a7a8a060bb3103abcf1c4e509f72d6b324384d71/meshmode/dof_array.py#L359-L378). Would that do the trick?

TIL 🙂 Yes, I think it might. On the other hand, this change could remove the existing code duplication (i.e., `rec_map_dof_array_container` could be reimplemented in terms of `rec_map_array_container`). Either solution would seem to work for me.

arraycontext/container/traversal.py
majosm: Whoops! Thanks, I'll fix it.

alexfikl:
> On the other hand, this change could remove the existing code duplication (i.e., `rec_map_dof_array_container` could be reimplemented in terms of `rec_map_array_container`). Either solution would seem to work for me.

It actually was initially! Well, in terms of `_map_array_container_impl`. I re-implemented it directly in `meshmode` at some point in the hopes that it would be a bit faster (and not too much of a maintainence burden, since the code is straightforward).

I don't feel particularly strongly about it either, so we can let @inducer break the tie :grin:

inducer:
I'm gravitating towards having this implemented here and making the `meshmode.dof_array` use this via the `leaf_class` argument. I don't see too much benefit in keeping both virtually-identical implementations around:
* https://github.com/inducer/meshmode/blob/a7a8a060bb3103abcf1c4e509f72d6b324384d71/meshmode/dof_array.py#L359-L378
* https://github.com/inducer/arraycontext/blob/957be2f5fb234bd3bf6a15c83099ba7b1e5bcb25/arraycontext/container/traversal.py#L90-L101

I also feel like the speed difference is probably non-existent. (pending benchmarks proving me wrong :grin: )

alexfikl:
> I also feel like the speed difference is probably non-existent. (pending benchmarks proving me wrong)

Made a simple benchmark using [this patch](https://gist.github.com/alexfikl/a4d5280270badd31bb18375e2c4394c8) and the version from `meshmode` is a bit slower!
```
INFO:__main__:[mm] min 6.767774e-03 mean 6.942166e-03 ± 1.912915e-04
INFO:__main__:[ac] min 5.201670e-03 mean 5.289515e-03 ± 7.113123e-05
```
so mostly just confused now :confused:

arraycontext/container/traversal.py
majosm: Thoughts on b395fa827236cb9028fe85089b8dd7941b500630? Things ended up getting a little hairy with this; in order to preserve backwards compatibility and not have to add the `()` when not using `leaf_class=`, `[multi]mapped_over_array_containers` now behaves either as a decorator or decorator factory depending on how it's called. Not sure if this is OK?

alexfikl: Yeah, looks good to me! At least I don't know of a cleaner way to do this sort of stuff.

Do these have a test to make sure they work in both cases?

majosm: Doesn't look like it. I added some in 442a119f5a126cb31de5fabdfe2175da71b40b07.

majosm: Doesn't look like it. I added some in 442a119f5a126cb31de5fabdfe2175da71b40b07.

alexfikl: That's great! Thank you!

majosm:
Ready for another look @inducer.

inducer (approved):
LGTM, thanks! If we can ever get the (currently super flaky) CI to pass, this'll land.

PR #129, opened 2021-12-15
Allow empty sub-containers in reductions

majosm:
Allows some (but not all) of the sub-containers in an array container to be empty in reductions. This allows reductions to be used on single-species CV objects in mirgecom.

Depends on #128 (not in an essential way, just avoiding dealing with merge conflicts).


inducer:
This has picked up some merge conflicts. Could you take a look? Also, ready to un-draft?

arraycontext/container/traversal.py
```diff
@@ -473,11 +473,22 @@ def rec(_ary: ArrayOrContainerT) -> ArrayOrContainerT:
...
+                if len(filtered_subary_results) > 0:
+                    return reduce_func(filtered_subary_results)
+                else:
+                    return None
 
-    return rec(ary)
+    result = rec(ary)
+
+    if result is None:
+        raise ValueError("cannot reduce empty array container")
```
inducer, line 489: Personally, I think reducing over an empty container is well-defined: just return the neutral element. `reduce_func` should be trusted to do the right thing. What was the reasoning for putting in this error message? Basically, what I'm pushing for is to get rid of it and return `reduce_func([])` instead.

arraycontext/container/traversal.py
```diff
@@ -503,12 +514,23 @@ def rec_multimap_reduce_array_container(
...
+        else:
+            return None
 
-    return _multimap_array_container_impl(
+    result = _multimap_array_container_impl(
         map_func, *args,
         reduce_func=_reduce_wrapper, leaf_cls=leaf_class, recursive=True)
 
+    if result is None:
+        raise ValueError("cannot reduce empty array container")
```
inducer, line 530: As above.

inducer (commented):
Thanks! LGTM aside from the all-empty handling, which I think ought to be quick to resolve.

arraycontext/container/traversal.py
majosm: If `reduce_func` worked for an empty list, I think this PR wouldn't be needed, right?

The motivation for this was for things like `actx.np.max` where `reduce_func` doesn't (currently) behave nicely for empty inputs. Should that maybe be fixed instead?

alexfikl: For reference, calling `np.max` on an empty list raises
```
>>> np.max([])
ValueError: zero-size array to reduction operation maximum which has no identity
```
and requires specifying an initial value with `np.max([], initial=np.inf)`. We should probably do something like that in `actx.np` too.

inducer: > The motivation for this was for things like `actx.np.max` where `reduce_func` doesn't (currently) behave nicely for empty inputs.

That's a great point. I had not thought far enough to realize that, but I agree now.

> Should that maybe be fixed instead?

I think that's exactly what we should do.

> We should probably do something like that in `actx.np` too.

I agree. While I wasn't loving numpy's interface of "no implicit neutral" element at first, I think it makes sense from the perspective of integers. I think we should aim to replicate this, both at the level of `actx.np`, and for the array container reductions under discussion here.

One subtlety is the `ValueError` for empty arrays. For symbolically-shaped `pytato` arrays, it may not be possible to do a perfect job.

inducer: Did some work towards that:
- https://github.com/inducer/pytato/pull/238
- https://github.com/inducer/arraycontext/pull/136

That latter PR needs some more work done to it. @majosm, could you take a look?

inducer:
@majosm, depending on how you feel, #136 might supersede this PR.

PR #131, opened 2021-12-16
Fix type mismatch in serialize_container argument

### LINKED ISSUE #130: CI failing in mypy
Failing CI run [here](https://github.com/inducer/arraycontext/runs/4551303186?check_suite_focus=true). Any ideas?

Seems to have started occurring today (CI runs in #129 passed yesterday).

alexfikl:
Seems like the new version of mypy is looking into the `singledispatch` arguments a bit more closely. The issue was that
* the base `serialize_container` took an `ArrayContainer` (which is an actual class)
* the `ndarray` version took an `ndarray` (which is not a subclass of `ArrayContainer`!)

This just sets it to `Any` so that it's a proper fallback for all objects.

Fixes #130

### DIFF
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -122,7 +122,7 @@ class NotAnArrayContainerError(TypeError):
 
 
 @singledispatch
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

PR #132, opened 2021-12-17
einsum: demote warning to comment

matthiasdiener:
This warning creates a lot of noise when running mirgecom


kaushikcfd:
Now that the Loopy code-generator is smarter in `pytato`, we could make use of: https://documen.tician.de/pytato/array.html#pytato.tags.PrefixNamed.

inducer:
> Now that the Loopy code-generator is smarter in `pytato`, we could make use of: https://documen.tician.de/pytato/array.html#pytato.tags.PrefixNamed.

I agree that that's better. Implemented in #135, which I think supersedes this. That way, if einsum arguments end up materialized (which they often might be), these names won't be lost.

PR #136, opened 2021-12-28
Match numpy interface for empty reductions

inducer:
- [ ] Needs https://github.com/inducer/pytato/pull/238
- [ ] Needs equivalent support for the pyopencl array context (@majosm, could you take that on?)
- [ ] Needs some tests (@majosm, could you take that on?)
- [ ] `requirements.txt` modification needs to be undone

cc @majosm


arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -91,22 +95,50 @@ def minimum(self, x, y):
...
+                ary)
+
+    # * appears where positional signature starts diverging from numpy
+    def sum(self, a, axis=None, dtype=None, *, initial=0):
+        import operator
+        return self._reduce(operator.add, pt.sum, a,
+                axis=axis, dtype=dtype, initial=initial)
+
+    # * appears where positional signature starts diverging from numpy
+    def min(self, a, axis=None, *, initial=_NoValue):
```
majosm, line 134: How come these need `_NoValue` instead of just `None`?

arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -91,22 +95,50 @@ def minimum(self, x, y):
...
+            ary, *,
+            axis, dtype, initial):
+        def container_reduce(ctr):
+            if initial is _NoValue:
+                try:
+                    return reduce(container_binop, ctr)
+                except TypeError as exc:
+                    assert "empty sequence" in str(exc)
+                    raise ValueError("zero-size reduction operation "
+                            "without supplied 'initial' value")
```
majosm, line 109: Should this maybe be
```suggestion
                    if "empty sequence" in str(exc):
                        raise ValueError("zero-size reduction operation "
                                "without supplied 'initial' value")
                    else:
                        raise        
```
? (In case an unrelated `TypeError` is raised.)

majosm (commented):
Just a couple of things I was wondering about while attempting to add the PyOpenCL equivalents.

arraycontext/impl/pytato/fake_numpy.py
inducer: Because you could ask for `None` to be returned in the case of an empty container.

PR #138, opened 2022-01-11
Tighten type info on `{rec_,}keyed_map_array_container`

inducer:
Not particularly urgent, but more informative IMO.

cc @matthiasdiener


kaushikcfd (approved):
LGTM!

PR #142, opened 2022-01-13
Add a leaf_class parameter to flatten

alexfikl:


arraycontext/container/traversal.py
```diff
@@ -629,12 +632,36 @@ def _flatten(subary: ArrayOrContainerT) -> None:
...
+            return actx.np.concatenate(result)
+
+    def _flatten_with_leaf_class(subary: ArrayOrContainerT) -> Any:
+        if type(subary) is leaf_class:
+            return _flatten_without_leaf_class(subary)
+
+        try:
+            iterable = serialize_container(subary)
+        except NotAnArrayContainerError:
+            return subary
```
alexfikl, line 654: Should this do a `np.ravel(subary)` so that all the arrays are 1d?

alexfikl:
This is also getting a lot more usage at
https://github.com/inducer/pytential/actions/runs/1693575089
and seems to work just fine!

arraycontext/container/traversal.py
```diff
@@ -583,7 +583,10 @@ def thaw(ary: ArrayOrContainerT, actx: ArrayContext) -> ArrayOrContainerT:
 
 # {{{ flatten / unflatten
 
-def flatten(ary: ArrayOrContainerT, actx: ArrayContext) -> Any:
+def flatten(
+        ary: ArrayOrContainerT, actx: ArrayContext, *,
+        leaf_class: Optional[type] = None,
```
inducer, line 588: Document `leaf_class`.

arraycontext/container/traversal.py
```diff
@@ -624,17 +626,40 @@ def _flatten(subary: ArrayOrContainerT) -> None:
...
+    def _flatten_without_leaf_class(subary: ArrayOrContainerT) -> Any:
+        result = _flatten(subary)
+
+        if len(result) == 1:
+            return result[0]
+        else:
+            return actx.np.concatenate(result)
+
+    def _flatten_with_leaf_class(subary: ArrayOrContainerT) -> Any:
+        if type(subary) is leaf_class:
```
inducer, line 646: Exact match or `isinstance`? What's the precedent from other `leaf_class`es? Also: document.

arraycontext/container/traversal.py
```diff
@@ -624,17 +626,40 @@ def _flatten(subary: ArrayOrContainerT) -> None:
...
             for _, isubary in iterable:
-                _flatten(isubary)
+                result.extend(_flatten(isubary))
 
-    _flatten(ary)
+        return result
 
-    if len(result) == 1:
-        return result[0]
+    def _flatten_without_leaf_class(subary: ArrayOrContainerT) -> Any:
```
inducer, line 637: Not sure this needs to be a separate function. Maybe just bolt this to the end of `flatten`?

arraycontext/container/traversal.py
inducer, line 588: Does `unflatten` just do the right thing when run on a with-leaf-class-flattened thing? If so, that might be a good thing to also explicitly document.

inducer, line 588: I'm not sure `leaf_class` is a great name. `flatten_at_class`?

inducer (commented):
Thanks! A few wrinkles aside, this is looking great.

inducer:

arraycontext/container/traversal.py
alexfikl: @inducer Mostly went with what was used in `rec_map_array_container` and friends in #128, since I would expect them to all use the same argument name. Would you prefer that they each have their own names like this?

arraycontext/container/traversal.py
alexfikl: All the other `rec_map` things don't use `isinstance` either. I'm not sure which choice would be better, since we mostly just have `DOFArray`s at the moment..

arraycontext/container/traversal.py
alexfikl: No, `unflatten` doesn't really work with this "mode" of `flatten`. Not sure if it's worth implementing that, since it's not really used anywhere at the moment.

arraycontext/container/traversal.py
alexfikl: Made it into a function because it's also called when `type(subary) is leaf_class`, mostly to just not repeat the `if len(result) == 1` part.

arraycontext/container/traversal.py
alexfikl: Added some docs in 5e87623.

Should also do the same for all the other functions using `leaf_class`, since none of them had some nice docs I could copy-paste :cry:

arraycontext/container/traversal.py
alexfikl: Mentioned in 5e87623 that subclasses are not considered.

arraycontext/container/traversal.py
alexfikl: Added a sentence to say that `unflatten` will not roundtrip if `leaf_class` is given in 5e87623.

alexfikl: Make an issue of it?

inducer: Ah :shrug:. Terminology shmerminology.

arraycontext/container/traversal.py
inducer: We're "above" the leaf class here, so I'd say it's consistent to leave the arrays alone.

PR #143, opened 2022-01-16
Add numpy support in each array context to the docs

alexfikl:
This does two things:
1. Rearranges the functions in `fake_numpy.py` files to match the sections and order in the (current?) numpy docs.
2. Adds a little script that automatically generates what each array context actually supports.

Unfortunately the first change moves a lot of code around and looks horrible :(

### DIFF
--- a/arraycontext/impl/pytato/fake_numpy.py
+++ b/arraycontext/impl/pytato/fake_numpy.py
@@ -54,96 +54,45 @@ class PytatoFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
         :mod:`pytato` does not define any memory layout for its arrays. See
         :ref:`Pytato docs <pytato:memory-layout>` for more on this.
     """
+
+    _pt_funcs = frozenset({
+        "sin", "cos", "tan", "arcsin", "arccos", "arctan",
+        "sinh", "cosh", "tanh", "exp", "log", "log10",
+        "sqrt", "abs", "isnan"
+        })
+
     def _get_fake_numpy_linalg_namespace(self):
         return PytatoFakeNumpyLinalgNamespace(self._array_context)
 
     def __getattr__(self, name):
-
-        pt_funcs = ["abs", "sin", "cos", "tan", "arcsin", "arccos", "arctan",
-                    "sinh", "cosh", "tanh", "exp", "log", "log10", "isnan",
-                    "sqrt", "exp"]
-        if name in pt_funcs:
+        if name in self._pt_funcs:
             from functools import partial
             return partial(rec_map_array_container, getattr(pt, name))
 
         return super().__getattr__(name)
 
+    # NOTE: the order of these follows the order in numpy docs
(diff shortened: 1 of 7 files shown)


alexfikl:
Looks something like this

Screenshot_20220115_204659

arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -51,96 +51,47 @@ class PytatoFakeNumpyNamespace(BaseFakeNumpyNamespace):
...
-                    "sinh", "cosh", "tanh", "exp", "log", "log10", "isnan",
-                    "sqrt", "exp"]
-        if name in pt_funcs:
+        if name in self._pt_funcs:
             from functools import partial
             return partial(rec_map_array_container, getattr(pt, name))
+        else:
+            raise AttributeError(
+                    f"'{type(self._array_context).__name__}.np' object "
+                    f"has no attribute '{name}'")
```
alexfikl, line 71: This is an actual functionality change! Before it would fallback to the default loopy generation code, but it doesn't seem like that was intended?

inducer:
Sorry, this picked up a few merge conflicts. :grimacing: Could you take a look?

alexfikl:
> Sorry, this picked up a few merge conflicts. :grimacing: Could you take a look?

Should be clean now!

doc/make_numpy_coverage_table.py
```diff
@@ -0,0 +1,247 @@
...
+            numpy_functions_for_context=build_supported_functions(funcs, contexts),
+            )
+    outf.write(r)
+
+
+def write_mathematical_functions(outf, contexts):
+    # https://numpy.org/doc/stable/reference/routines.math.html
+    funcs = (
+            ("data", "sin"),
+            ("data", "cos"),
```
alexfikl, line 173: A lot of the functions listed in [_numpy_math_functions](https://github.com/inducer/arraycontext/blob/5218b2d6398d818f517d3a9856a9dabb6076dea0/arraycontext/fake_numpy.py#L78-L129) don't actually work. They error out with something like
```
loopy.diagnostic.LoopyError: could not determine type of 'out' of kernel 'actx_special_sinc'.
```

inducer: Those are just not available in loopy then. Not ideal, but that's life for now.

inducer:
Thanks for doing this. This makes a good deal of sense to me.

doc/make_numpy_coverage_table.py
alexfikl: Wouldn't it be better to comment them out in `_numpy_math_functions` so that we get the nicer `AttributeError`? That `LoopyError` traceback is very scary!

inducer: I'd rather improve the error message in loopy. If we comment them out, then there are two sources of truth on what's supported. Otherwise there's only one, which IMO is preferable.

alexfikl: That does sound better, yeah.

PR #88, opened 2021-09-16
Removing BaseFakeNumpySpace's dependency on call_loopy

mitkotak:
Resolve [this](https://github.com/inducer/arraycontext/issues/70) issue by moving existing scalar unfunc logic into arraycontext.loopy in order to remove BaseFakeNumpySpace's dependency on call_loopy

PR #144, opened 2022-01-16
Remove stray _get_scalar_func_loopy_program

alexfikl:
Noticed this didn't get removed in #88 when merging it into #143.


inducer:
Thanks for catching that!

cc @mitkotak

kaushikcfd:
Thanks!

PR #147, opened 2022-02-20
Fix TaggableCLArray copy without queue

alexfikl:


arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -73,13 +73,7 @@ def copy(self, queue=cla._copy_queue, tags=None, axes=None, _new_class=None):
...
         """
         _new_class = self.__class__ if _new_class is None else _new_class
-
-        if queue is not cla._copy_queue:
-            # Copying command queue is an involved operation, use super-class'
-            # implementation.
-            base_instance = super().copy(queue=queue)
-        else:
-            base_instance = self
+        base_instance = super().copy(queue=queue)
```
alexfikl, line 76: @kaushikcfd I'm not sure what the intended effect was here, but it seems like in the else case it doesn't do a copy?

kaushikcfd: Yep, that was the idea. Did this run into any issues?

arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -125,5 +122,8 @@ def to_tagged_cl_array(ary: cla.Array,
         array. If passed *None*, then initialized to a :class:`pytato.Axis`
         with no tags attached for each dimension.
     """
+    if isinstance(ary, TaggableCLArray):
+        return ary
+
```
kaushikcfd, line 127: This slightly worries me that the tags passed to `to_tagged_cl_array` are ignored. OTOH, I see your point of how do we handle conflicts. How about we preserve both the tags (the ones present in `ary.tags` and those passed in as `tags` arguments) and clarify this behavior in the docs?

arraycontext/impl/pyopencl/taggable_cl_array.py
alexfikl: Yeah, when trying to do a proper copy of a `TaggableCLArray` with `ary.copy()` it doesn't actually do the copy, so any in-place operations change the same underlying data.

If this was intended to be more immutable, maybe we can disallow in-place changes?

arraycontext/impl/pyopencl/taggable_cl_array.py
alexfikl: Oh, yeah, sorry, I should mark this as a draft still. 

I'll look into transfering the tags.

arraycontext/impl/pyopencl/taggable_cl_array.py
kaushikcfd: > Yeah, when trying to do a proper copy of a TaggableCLArray with ary.copy() it doesn't actually do the copy, so any in-place operations change the same underlying data. However, the exact same code with ary.copy(queue=queue) works.

Aah I see, makes sense. Yep your approach looks correct to me. Thanks for the fix!

kaushikcfd: > If this was intended to be more immutable, maybe we can disallow in-place changes?

I don't see a good reason to diverge from the underlying `cla.Array` behavior. IMO, we should continue allowing in-place updates.

alexfikl: > Aah I see, makes sense. Yep your approach looks correct to me. Thanks for the fix!

Reworked this a bit 
* added a `wrap_cl_array` method that does the second half of the `copy` that just adds the tags
* Use `wrap_cl_array` instead of `copy` when an actual copy of the underlying data is not intended.
* `TaggableCLArray.copy` now calls the `super().copy` and then `wrap_cl_array` to add tags

It seems to be working as intended now. What do you think?

arraycontext/impl/pyopencl/taggable_cl_array.py
alexfikl: Handled the tags in `to_tagged_cl_array` in 2ac228f. Not super sure the `axes` are handled correctly, so let me know if anything is fishy there!

inducer:
So I think the underlying issue here is that there's confusion on what `copy` should do. (Which I should have caught before merging #111, sorry!)
- `Taggable` assumes `copy` can be used for (cheap-ish, immutable) return-with-modifications.
- `cla.Array` says `copy` copies array data.

IMO:
- It's not possible to cleanly untangle those two use cases without causing un-obvious corner cases.
- `TaggableCLArray` should not override `copy`, and `Taggable` should not assume `copy` is used to apply `tags`. The latter makes sense for another reason: If the `Taggable` is a dataclass, then the correct spelling would be something with `replace`.

So my vote would be to give `Taggable` an interface (`_with_tags`) that a subclass needs to define, with a deprecated fallback to using `copy`.

What do the both of you think?

alexfikl:
> What do the both of you think?

I think that's mostly what the current version does, except that `copy` still tries to handle tags. I can remove that once we agree on an approach.

For me this is :+1:

inducer:
https://github.com/inducer/pytools/pull/120 realizes part of this.

> EDIT: Not super worried about deprecating things though, since this was just introduced and nobody uses it outside `arraycontext` (right?).

I'm talking about changing `Taggable`, and that has uses throughout `pytato`, here, and `loopy`.

alexfikl:
> I'm talking about changing `Taggable`, and that has uses throughout `pytato`, here, and `loopy`.

Ah! I read that as a shorthand for `TaggableCLArray` and got confused. Then yeah, completely agree that needs a bit more care. Thanks for taking care of the `pytools` side!

alexfikl:
@inducer This needs a `pytools` release that includes inducer/pytools#120. Otherwise it should be good for another look!

inducer:
https://pypi.org/project/pytools/2022.1.1/

inducer:
Thx! Pushed a `pytools` version constraint to (hopefully) make mirgecom happy, and underscored `unwrap_cl_array`. LGTM, in it goes, pending CI. :)

alexfikl:
> Thx! Pushed a `pytools` version constraint to (hopefully) make mirgecom happy, and underscored `unwrap_cl_array`. LGTM, in it goes, pending CI. :)

Looks like it picked it up! Thank you.

PR #149, opened 2022-03-01
Make `get_container_context_recursively` return actx or error

inducer:
This is a deliberate compatibility break based on seeing code like this:

https://github.com/inducer/grudge/blob/62af257a45be631e2bf1d3c4cf9de4ec746f375d/grudge/trace_pair.py#L307

which clearly would not succeed if the return value was `None`, based on the notion that the original `get_container_context_recursively` is easy to misuse because of its `Optional` return value. The rest of this is  a renaming of `get_container_context` to `get_container_context_opt` for consistency.

As for the compatibility break, my justification is that all existing user code seems to be using the new `get_container_context_recursively` already. :)

### DIFF
--- a/arraycontext/__init__.py
+++ b/arraycontext/__init__.py
@@ -40,7 +40,8 @@ from .metadata import _FirstAxisIsElementsTag
 from .container import (
         ArrayContainer, NotAnArrayContainerError,
         is_array_container, is_array_container_type,
-        get_container_context, get_container_context_recursively,
+        get_container_context_opt,
+        get_container_context_recursively, get_container_context_recursively_opt,
         serialize_container, deserialize_container,
         register_multivector_as_array_container)
 from .container.arithmetic import with_container_arithmetic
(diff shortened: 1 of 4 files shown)


alexfikl:
If I understood correctly, the idea is that if someone calls `get_array_context_recursively` they must really want a context, so it should yell more loudly if it doesn't find one?

inducer:
Yep, that's the idea.

inducer:
Given the mirgecom failure, I think I might start by warning about `None` returns for now, patch up mirgecom, and then put the rest of this plan into action.

alexfikl:
> Given the mirgecom failure, I think I might start by warning about `None` returns for now, patch up mirgecom, and then put the rest of this plan into action.

Maybe add a `strict` flag like a few of the other functions and warn when people set it to a deprecated `strict=False`?

alexfikl (approved):

PR #150, opened 2022-03-17
Add `flat_size_and_dtype`

inducer:
This is needed for buffer allocation in MPI communication in (multi-volume) grudge. Otherwise, the only way to find these metadata of a flattened buffer is to flatten it, which is a waste.

cc @majosm

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -27,6 +27,7 @@ Flattening and unflattening
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: flatten
 .. autofunction:: unflatten
+.. autofunction:: flat_size_and_dtype
 
 Numpy conversion
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


arraycontext/context.py
```diff
@@ -334,7 +334,7 @@ def einsum(self, spec, *args, arg_names=None, tagged=()):
         return self.tag(tagged, out_ary)
 
     @abstractmethod
-    def clone(self):
+    def clone(self) -> "ArrayContext":
```
inducer, line 337: Whoops. This slipped in. It's unrelated, but I guess it doesn't hurt.

arraycontext/container/traversal.py
```diff
@@ -771,6 +772,35 @@ def _unflatten(template_subary: ArrayOrContainerT) -> ArrayOrContainerT:
 
     return result
 
+
+def flat_size_and_dtype(
+        ary: ArrayOrContainerT) -> Tuple[int, Optional[np.dtype[Any]]]:
+    """Return a tuple ``(size, dtype)`` that would be the length and
+    :class:`numpy.dtype` of the one-dimensional array returned by
+    :func:`flatten`.
```
alexfikl, line 780:
```suggestion
    """
    :returns: a tuple ``(size, dtype)`` that would be the length and
    	:class:`numpy.dtype` of the one-dimensional array returned by
    	:func:`flatten`.
```
?

alexfikl (approved):
Looks good to me!

inducer:
Thanks for taking a look!

PR #153, opened 2022-04-05
Allow dataclass containers with DeviceArrays

alexfikl:
This essentially just made `DeviceArray` and `DeviceScalar` into distinct `Protocol` classes (not just `DeviceScalar = Any`) and added a little check in `dataclass_array_container`.

The motivation is to maybe eventually replace `boxtree.DeviceDataRecord` with some standard dataclasses that support `arraycontext`.

### DIFF
--- a/arraycontext/context.py
+++ b/arraycontext/context.py
@@ -75,18 +75,8 @@ The interface of an array context
 
 .. currentmodule:: arraycontext
 
-.. class:: DeviceArray
-
-    A (type alias for an) array type supported by the :class:`ArrayContext`
-    meant to aid in typing annotations. For a explicit list of supported types
-    see :attr:`ArrayContext.array_types`.
-
-.. class:: DeviceScalar
-
-    A (type alias for a) scalar type supported by the :class:`ArrayContext`
-    meant to aid in typing annotations, e.g. for reductions. In :mod:`numpy`
-    terminology, this is just an array with a shape of ``()``.
-
+.. autoclass:: Array
+.. autoclass:: Scalar
 .. autoclass:: ArrayContext
 """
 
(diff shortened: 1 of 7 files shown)


arraycontext/context.py
```diff
@@ -123,10 +113,55 @@
...
+    """A :class:`~typing.Protocol` for the scalar type supported by
+    :class:`ArrayContext`.
+
+    In :mod:`numpy` terminology, this is just an array with a shape of ``()``.
+
+    This is meant to aid in typing annotations. For a explicit list of
+    supported types see :attr:`ArrayContext.array_types`.
+
+    .. attribute:: shape
+    .. attribute:: dtype
```
inducer, line 156: `dtype` missing as a property?

arraycontext/context.py
```diff
@@ -123,10 +113,55 @@
...
+
 _ScalarLike = Union[int, float, complex, np.generic]
 
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol                  # type: ignore[misc]
+
+
+class DeviceArray(Protocol):
```
inducer, line 126: Maybe use this in the type annotations for the containers? (That's undoubtedly a bigger project, so we can alway tackle that later.)

inducer, line 126: Suppose this is our last change to discuss the name. Is `DeviceArray` the best we can do? It's kind of a poor fit for many of the use cases. (I know there's a precedent, but we *could* change/deprecate this before we plaster this name absolutely everywhere.)

How about just `Array`? Or `ContextArray` (:vomiting_face:)?

arraycontext/context.py
```diff
@@ -123,10 +113,55 @@
...
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        ...
+
+    @property
+    def dtype(self) -> "np.dtype[Any]":
+        ...
+
+
+class DeviceScalar(DeviceArray, Protocol):
```
inducer, line 146: Similar naming concern as above. Just `Scalar`?

inducer (commented):
Thanks! I sympathize with the goal of getting boxtree less hooked on CL arrays.

arraycontext/context.py
alexfikl: Hm, I'm not too bothered by `DeviceArray`. That's what JAX calls its CPU/GPU array class and `Device` makes sense in the `pyopencl` case too. I imagine it's a worse fit for the lazy case?

I wouldn't feel bad about just calling it `Array` either (as a component of an `ArrayContainer`). @kaushikcfd Thoughts or suggestions for better naming?

arraycontext/context.py
alexfikl: It supposedly inherits from `DeviceArray`, so should have it. 

Apparently you can't make a `Protocol` runtime checkable if it doesn't have any actual methods, so I guess the inheritance is pretty useless..

arraycontext/context.py
kaushikcfd: `Array` or `ArrayT` sound better, compared to the other options here :). (numpy arrays are also not quite "device arrays")

alexfikl: I think most of the code in `arraycontext.container.traversal` already uses it a bit. I have another branch/PR on top of this that adds more annotations to `ArrayContext` using these, so we can give it a more thorough look there.

alexfikl: Renamed to `Array` in https://github.com/inducer/arraycontext/compare/af65ccc1c8b6edbb2d454757a726ef0cd3484431..e7a0fd3972ba3d794ffd0d656b3e736fa42d1392

arraycontext/context.py
```diff
@@ -123,10 +113,64 @@
...
+        ...
+
+    @property
+    def dtype(self) -> "np.dtype[Any]":
+        ...
+
+
+# NOTE: backwards compatibility
+DeviceArray = Array
+DeviceScalar = Scalar
```
inducer, line 170: Should we do a deprecation via `__getattr__` on these?

alexfikl: From a quick grep, only `DeviceScalar` is used in a few places in grudge.

Added in https://github.com/inducer/arraycontext/compare/a9252057a392bcab1d0ac79ac193e9040a60d6f9..160c66dd044932e698641e57d9f9197d239e252e.

inducer:
Looks great. Thanks for working on this!

PR #155, opened 2022-04-13
`with_container_arithmetic`: Default cls_has_array_context_attr based on presence of attribute

inducer:
I think it makes sense for `WaveState * pytato.Array` to just work, but currently it won't, because `with_container_arithmetic` will only access `.array_context` on a container (to get the array types) if `_cls_has_array_context_attr` is set to `True`. Just defaulting it to True if the attribute exists also doesn't work, because many of those properties are implemented like this:
https://github.com/inducer/arraycontext/blob/b75ba4f60c601664edc495cc3606fc9f7e47866b/test/test_arraycontext.py#L684-L686=
but are then used with numpy arrays in the `.mass` attribute, where that implementation will fail.

*If* a class has the `.array_context` attribute, the code in this PR tries harder to get a hold of the array context involved, ultimately nudging people to fix their `.array_context` property implementations and passing `_cls_has_array_context_attr=True` (which we may then deprecate even further into the future, using just existence of the attribute instead).

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -31,7 +31,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Any, Callable, Optional, Tuple, TypeVar, Union
+from warnings import warn
+from typing import Any, Callable, Optional, Tuple, TypeVar, Union, Type
 
 import numpy as np
 
@@ -124,6 +125,10 @@ def _format_binary_op_str(op_str: str,
         return op_str.format(arg1, arg2)
 
 
+class _FailSafe:
+    pass
+
+
 def with_container_arithmetic(
         *,
         bcast_number: bool = True,
@@ -135,7 +140,7 @@ def with_container_arithmetic(
         matmul: bool = False,
         bitwise: bool = False,
         shift: bool = False,
-        _cls_has_array_context_attr: bool = False,
+        _cls_has_array_context_attr: Optional[bool] = None,
         eq_comparison: Optional[bool] = None,
         rel_comparison: Optional[bool] = None) -> Callable[[type], type]:
     """A class decorator that implements built-in operators for array containers
@@ -172,6 +177,8 @@ def with_container_arithmetic(
         class has an ``array_context`` attribute. If so, and if :data:`__debug__`
         is *True*, an additional check is performed in binary operators
         to ensure that both containers use the same array context.
+        If *None* (the default), this value is set based on whether the class
+        has an ``array_context`` attribute.
         Consider this argument an unstable interface. It may disappear at any moment.
 
     Each operator class also includes the "reverse" operators if applicable.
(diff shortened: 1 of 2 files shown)


inducer:
@alexfikl Sorry about the review request... this clearly needs some more thought.

arraycontext/container/arithmetic.py
```diff
@@ -260,6 +257,42 @@ def numpy_pred(name: str) -> str:
...
+                        "deprecated and will stop working in 2023. "
+                        "Having a working .array_context attribute is desirable "
+                        "to enable arithmetic with other array types supported "
+                        "by the array context."
+                        f"If {cls}.array_context will not fail, pass "
+                        "_cls_has_array_context_attr=True. "
+                        "If you do not want container arithmetic to make "
+                        "use of the array context, set "
+                        "_cls_has_array_context_attr=False.",
+                        stacklevel=2)
```
alexfikl, line 282: `else: cls_has_array_context_attr = False`?

alexfikl: Ah, nevermind, it doesn't add these lower down!

test/test_arraycontext.py
```diff
@@ -672,36 +672,44 @@ def test_array_context_einsum_array_tripleprod(actx_factory, spec):
...
 @with_container_arithmetic(bcast_obj_array=False,
-        eq_comparison=False, rel_comparison=False)
+        eq_comparison=False, rel_comparison=False,
+        _cls_has_array_context_attr=True)
 @dataclass_array_container
 @dataclass(frozen=True)
 class MyContainer:
     name: str
-    mass: DOFArray
+    mass: DOFArray   # or np.ndarray
```
alexfikl, line 681: We need some work in `dataclass_array_container` to make this work with a `Union`, right?

alexfikl (approved):
This is annoyingly complicated, but yeah, can't think of a nicer way to do it so :+1:

inducer:
Thanks for taking a look!

> This is annoyingly complicated

I totally agree. I kind of hate it. But at least it puts us on a course *towards* more sanity? At some point? Maybe?

arraycontext/container/arithmetic.py
```diff
@@ -268,18 +301,66 @@ def wrap(cls: Any) -> Any:
...
+                        "deprecated. "
+                        "Please fix the implementation of "
+                        f"'{type(ary).__name__}.array_context' "
+                        "and then set _cls_has_array_context_attr=True when "
+                        "calling with_container_arithmetic to avoid the run time "
+                        "cost of the check that gave you this warning. "
+                        "Using expensive recovery for now.",
+                        DeprecationWarning, stacklevel=3)
+
+                return get_container_context_recursively(ary)
```
alexfikl, line 341: Would it make sense to cache this in a `self.__actx` or something so it doesn't have to redo `get_container_context_recursively` on every math operation?

(same for `_get_actx_array_types_failsafe`)

alexfikl: Although that's the very pessimistic case, so might not be worth adding some more complexity in here..

inducer: IMO: No. The slower and noisier this is, the better, because that means we can get rid of all this nonsense sooner. There's an easy way out. Just set `_cls_has_array_context_attr=True` and none of this ever happens.

test/test_arraycontext.py
inducer: I seem to recall we tried, and it wasn't easy.

inducer:
OK, what's the worst that could happen. Let's land it! :) :tada:

PR #156, opened 2022-04-20
Support unions in dataclass_array_container

alexfikl:
Something like this?


inducer:
Nice, thanks! I guess I was reluctant to start picking apart unions, but looking back at that thought process, it doesn't make a lot of sense. The structure is documented and unlikely to change, after all.

PR #158, opened 2022-05-01
PytatoArrayContext.freeze: support container freezes

kaushikcfd:
Freezing across containers can potentially re-use common sub-expressions.

- :warning: Needs to go in conjunction with https://github.com/inducer/meshmode/pull/327.

Draft because:
- [x] Pass CI by correctly pointing the meshmode branch in the downstream CI.
- [x] DROP final commit before merge

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -880,7 +880,10 @@ def to_numpy(ary: ArrayOrContainerT, actx: ArrayContext) -> Any:
                     f"array of type '{type(subary).__name__}' not in "
                     f"supported types {actx.array_types}")
 
-    return rec_map_array_container(_to_numpy_with_check, ary)
+    return rec_map_array_container(_to_numpy_with_check,
+                                   # do a freeze first, if 'actx' supports
+                                   # container-wide freezes
+                                   thaw(freeze(ary, actx), actx))
 
 # }}}
 
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -65,13 +65,15 @@ THE SOFTWARE.
 
 from typing import Any, Callable, Iterable, List, Optional, Union, Tuple
 from functools import update_wrapper, partial, singledispatch
+from warnings import warn
 
 import numpy as np
 
 from arraycontext.context import ArrayContext, Array, _ScalarLike
 from arraycontext.container import (
         ArrayT, ContainerT, ArrayOrContainerT, NotAnArrayContainerError,
-        serialize_container, deserialize_container)
+        serialize_container, deserialize_container,
+        get_container_context_recursively_opt)
 
 
 # {{{ array container traversal helpers
@@ -519,7 +521,6 @@ def rec_multimap_reduce_array_container(
 
 # {{{ freeze/thaw
 
-@singledispatch
 def freeze(
         ary: ArrayOrContainerT,
         actx: Optional[ArrayContext] = None) -> ArrayOrContainerT:
@@ -533,23 +534,33 @@ def freeze(
 
     See :meth:`ArrayContext.thaw`.
     """
-    try:
-        iterable = serialize_container(ary)
-    except NotAnArrayContainerError:
-        if actx is None:
-            raise TypeError(
-                    f"cannot freeze arrays of type {type(ary).__name__} "
-                    "when actx is not supplied. Try calling actx.freeze "
-                    "directly or supplying an array context")
-        else:
-            return actx.freeze(ary)
+
+    if actx is None:
+        warn("Calling freeze(ary) without specifying actx is deprecated, explicitly"
+             " call actx.freeze(ary) instead. This will stop working in 2023.",
+             DeprecationWarning, stacklevel=2)
+
+        actx = get_container_context_recursively_opt(ary)
     else:
-        return deserialize_container(ary, [
-            (key, freeze(subary, actx=actx)) for key, subary in iterable
-            ])
+        warn("Calling freeze(ary, actx) is deprecated, call actx.freeze(ary)"
+             " instead. This will stop working in 2023.",
+             DeprecationWarning, stacklevel=2)
+
+        if __debug__:
+            rec_actx = get_container_context_recursively_opt(ary)
+            if (rec_actx is not None) and (rec_actx is not actx):
+                raise ValueError("Supplied array context does not agree with"
+                                 " the one obtained by traversing 'ary'.")
+
+    if actx is None:
+        raise TypeError(
+                f"cannot freeze arrays of type {type(ary).__name__} "
+                "when actx is not supplied. Try calling actx.freeze "
+                "directly or supplying an array context")
+
+    return actx.freeze(ary)
 
 
-@singledispatch
 def thaw(ary: ArrayOrContainerT, actx: ArrayContext) -> ArrayOrContainerT:
     r"""Thaws recursively by going through all components of the
     :class:`ArrayContainer` *ary*.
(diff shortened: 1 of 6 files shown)


arraycontext/container/traversal.py
```diff
@@ -533,6 +533,9 @@ def freeze(
 
     See :meth:`ArrayContext.thaw`.
     """
+    if actx is not None and actx.supports_freezing_containers:
```
inducer, line 536: We should just deprecate these then (and expect all array contexts to support container freezes).

arraycontext/container/traversal.py
```diff
@@ -842,7 +845,7 @@ def _to_numpy_with_check(subary: Any) -> Any:
                     f"array of type '{type(subary).__name__}' not in "
                     f"supported types {actx.array_types}")
 
-    return rec_map_array_container(_to_numpy_with_check, ary)
+    return rec_map_array_container(_to_numpy_with_check, freeze(ary, actx))
```
inducer, line 848: Is this correct? (Calling `freeze` for `to_numpy`?)

arraycontext/container/traversal.py
inducer, line 536: `actx.thaw` should then also support containers, for symmetry.

kaushikcfd: Given that the freeze logic is specific to the containers (freeze is a single-dispatch function), deprecating this won't be possible.

arraycontext/container/traversal.py
kaushikcfd: Yep, was incorrect. Replaced with `thaw(freeze(...))`. See 98776255f9cf518172bc57a35d75380f948e6ef0.

arraycontext/container/traversal.py
kaushikcfd: Given that we are not replacing `freeze(..., actx)` with `actx.freeze`, I don't think we would be quite preserving symmetry here.

alexfikl: This should also be added to [meshmode.dof_array](https://github.com/inducer/meshmode/blob/b795b77d49422da2da90f1c0b07668d2723ae048/meshmode/dof_array.py#L381-L389) then?

alexfikl: As a side note, I really don't recall why we made this `singledispatch`-ed and can't think of a reason for it now. Anyone remember something?

kaushikcfd: >  I really don't recall why we made this singledispatch-ed and can't think of a reason for it now. Anyone remember something?

IIUC, array container initialization might be specific to the container type, and instead of forcing any constructor API we ask the container's writer to also provide a freeze. For example, DOFArray takes in a `.array_context` argument during initialization that must be set to `None` on getting frozen.

> This should also be added to [meshmode.dof_array](https://github.com/inducer/meshmode/blob/b795b77d49422da2da90f1c0b07668d2723ae048/meshmode/dof_array.py#L381-L389) then?

Yep!

alexfikl: > IIUC, array container initialization might be specific to the container type, and instead of forcing any constructor API we ask the container's writer to also provide a freeze. For example, DOFArray takes in a `.array_context` argument during initialization that must be set to `None` on getting frozen.

And using `deserialize_container` doesn't work because the template arg isn't necessarily frozen. Hm.. annoying..

Thanks for the reminder!

inducer:
Discussed with @alexfikl a bit regarding the issue of how to set `DOFArray.array_context` to `None` as part of a context-driven `freeze`. While, at the time, we agreed that we should get rid of that attribute entirely (discussion in https://github.com/inducer/arraycontext/issues/162), but that's a total non-starter in the short term for the amount of breakcage it causes (and possibly in the long term, too).

So the second-best solution we came up with is to deprecate `arraycontext.container.traversal.{freeze,thaw}` and make them aliases of the respective methods in the array context (which are now defined to allow containers). In addition, introduce a new `@singledispatch` interface `arraycontext.container.with_array_context` (used by the array context's `freeze` and `thaw`) that returns a new array with the corresponding array context stored inside of it.

How does that sound?

kaushikcfd:
> How does that sound?

Yep, this sounds like a cleaner approach. I've implemented the suggestions, however back-compatibility seems challenging (unachievable?). For ex. meshmode needs https://github.com/inducer/meshmode/pull/327.

inducer:
> back-compatibility seems challenging (unachievable?).

Yep, that's understood. But IMO `DOFArray` is the only container affected (for now), so I think that's OK. This is fixing  (papering over?) a genuine design mistake, so I'd rather deal with it than be locked into a dumb interface forever.

arraycontext/container/traversal.py
inducer: Fair enough.

arraycontext/container/traversal.py
```diff
@@ -533,20 +534,22 @@ def freeze(
...
-        iterable = serialize_container(ary)
-    except NotAnArrayContainerError:
-        if actx is None:
-            raise TypeError(
-                    f"cannot freeze arrays of type {type(ary).__name__} "
-                    "when actx is not supplied. Try calling actx.freeze "
-                    "directly or supplying an array context")
-        else:
-            return actx.freeze(ary)
+    if actx is None:
```
inducer, line 537: We should be able to make `freeze` not `@singledispatch` here then, right?

arraycontext/container/traversal.py
```diff
@@ -570,14 +573,35 @@ def thaw(ary: ArrayOrContainerT, actx: ArrayContext) -> ArrayOrContainerT:
         in :mod:`meshmode`. This was necessary because
         :func:`~functools.singledispatch` only dispatches on the first argument.
     """
+    if __debug__:
```
inducer, line 576: Likewise for `thaw`: I think that no longer needs `@singledispatch`.

arraycontext/container/traversal.py
inducer: Revisiting this: given the new logic, I think we ought to just deprecate the stand-alone `freeze` in favor of `actx.freeze`. (same for `thaw`) I don't see a strong rationale for keeping them.

arraycontext/container/traversal.py
```diff
@@ -570,14 +573,35 @@ def thaw(ary: ArrayOrContainerT, actx: ArrayContext) -> ArrayOrContainerT:
...
+# {{{ with_array_context
+
+@singledispatch
+def with_array_context(ary: ArrayOrContainerT,
+                       actx: Optional[ArrayContext]) -> ArrayOrContainerT:
+    """
+    Recursively associates *actx* to all the components of *ary*.
+
+    Array container types may use :func:`functools.singledispatch` ``.register``
+    to register additional implementations.
```
inducer, line 594:
```suggestion
    to register container-specific implementations.
    See `this issue <https://github.com/inducer/arraycontext/issues/162>`__
    for discussion of the future of this functionality.
```

arraycontext/container/traversal.py
kaushikcfd: Yep, removed!

arraycontext/container/traversal.py
kaushikcfd: Yep, removed! See https://github.com/inducer/arraycontext/compare/d373c46d46254bc539140152503c834a4fd3d441..cbd3adea6b48be9ca3d384106ba60125734e3e7f

arraycontext/container/traversal.py
inducer: In view of #162, I'd prefer to reduce our dependence on these, so deprecating them now seems like a good move. I know that this is will be super noisy on the application end, but I think it's a necessary (if painful) step.

I'd welcome feedback though. (cc @mtcam)

inducer: It looks like this has already happened for pytato and pyopencl, but not for Jax. Make it consistent?

inducer (commented):
Thanks! A few things to do here, but hopefully nothing onerous or disruptive. LGTM once these are take care of.

arraycontext/container/traversal.py
kaushikcfd: Thanks, done!

arraycontext/container/traversal.py
kaushikcfd: EagerJax and PytatoJax already support `actx.(freeze|thaw)` on containers. Anything I'm missing?

inducer: Sorry, misread this code. You're right!

kaushikcfd: Great! In that case, ok to orchestrate the merges of this andhttps://github.com/inducer/meshmode/pull/327?

inducer:
LGTM, thanks! Force-merging based on success of previous CI.

arraycontext/container/traversal.py
inducer: (Already on the way!)

PR #166, opened 2022-06-08
Deprecate passing unevaluated arrays to compiled functions

inducer:
This is potentially controversial, but I think the performance gotcha of accidentally evaluating stuff is considerable and worth the bother.

cc @mtcam

### DIFF
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -468,6 +468,14 @@ def _args_to_device_buffers(actx, input_id_to_name_in_program, arg_id_to_arg):
             pass
         elif isinstance(arg, pt.Array):
             # got an array expression => evaluate it
+            from warnings import warn
+            warn(f"Argument array '{arg_id}' to a compiled function is "
+                    "unevaluated. Evaluating just-in-time, at "
+                    "considerable expense. This is deprecated and will stop "
+                    "working in 2023. To avoid this warning, force evaluation "
+                    "of all arguments via freeze/thaw.",
+                    DeprecationWarning, stacklevel=4)
+
             arg = actx.freeze(arg)
         else:
             raise NotImplementedError(type(arg))


arraycontext/impl/pytato/compile.py
```diff
@@ -468,6 +468,14 @@ def _args_to_device_buffers(actx, input_id_to_name_in_program, arg_id_to_arg):
...
         elif isinstance(arg, pt.Array):
             # got an array expression => evaluate it
+            from warnings import warn
+            warn(f"Argument array '{arg_id}' to a compiled function is "
+                    "unevaluated. Evaluating just-in-time, at "
+                    "considerable expense. This is deprecated and will stop "
+                    "working in 2023. To avoid this warning, force evaluation "
+                    "of all arguments via freeze/thaw.",
+                    DeprecationWarning, stacklevel=4)
+
```
kaushikcfd, line 478: Should we support _thunking_ of pytato arrays instead? I.e. memoize the frozen values of a pytato array for a given target and during codegen short-circuit to use that value?

The reason I bring this up is because I'm imagining an RK4 function and with all the freeze/thaws it won't look pretty.

inducer: > Should we support _thunking_ of pytato arrays instead? I.e. memoize the frozen values of a pytato array for a given target and during codegen short-circuit to use that value?

Wrote something here: https://github.com/inducer/pytato/issues/232#issuecomment-1149392168. tl;dr: I'm thinking no.

> The reason I bring this up is because I'm imagining an RK4 function and with all the freeze/thaws it won't look pretty.

- In the short run, https://github.com/inducer/grudge/pull/252 obviates the need for that.
- In the longer term, I imagine we'll want to compile whole timesteps anyhow. Or, alternatively, timesteppers will be made by infrastructure a la leap.

In either case, the question doesn't arise.

inducer:
In a sense, the point of this PR is that we can't realistically afford any `freeze`s in a hot loop (such as a timestep loop).

arraycontext/impl/pytato/compile.py
MTCam: >... _thunking_ of pytato arrays ... 

🤯 who made this term up, please?

This PR seems like a great idea to me, but it hasn't knocked anything down yet.  I'm going to go try it out right now.

kaushikcfd: > exploding_head who made this term up, please?

Most probably the functional programming folks, probably early developers of Haskell.

> In either case, the question doesn't arise.

Fair enough. I do think this makes the performance transparent. Minor (non-blocking) recommendation: from 2023 allow taking an array, but make it raise only with `-O`.

kaushikcfd (approved):
LGTM, thanks!

arraycontext/impl/pytato/compile.py
inducer: > Minor (non-blocking) recommendation: from 2023 allow passing a `pt.Array`, but make it raise only with `-O`.

I don't know how I feel about that. I'm not comfortable announcing that now, but I might warm to it.

inducer:
@MTCam This'll probably be a deluge of warnings on your end, but I'd say they're typically worth knowing about. Let me know how bad it is. I'll hold off here until I hear back from you.

MTCam:
> @MTCam This'll probably be a deluge of warnings on your end, but I'd say they're typically worth knowing about. Let me know how bad it is. I'll hold off here until I hear back from you.

This is just initial, but I ran lots of tests with this last night using the prediction-like example, combozzle.   The warning did not trigger.   I will check out the other drivers and repair them with haste if they trigger this.   My .02: Merge with assumed impunity.

inducer:
Thanks all for providing feedback!

PR #167, opened 2022-06-09
[bugfix]: do not exit early for 0-size arrays

kaushikcfd:
In some cases, this would lose metadata.

Co-authored-by: Matthew Smith <mjsmith6@illinois.edu>

---

Originally proposed in https://github.com/kaushikcfd/arraycontext/pull/3 and makes sense now that pytato is smarter about empty arrays after https://github.com/inducer/pytato/pull/320.

### DIFF
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -296,17 +296,7 @@ class PytatoPyOpenCLArrayContext(_BasePytatoArrayContext):
                     raise TypeError(f"{type(self).__name__}.freeze invoked "
                                     f"with non-pytato array of type '{type(array)}'")
 
-                if subary.size == 0:
-                    # early exit for 0-sized arrays
-                    key_to_frozen_subary[key] = to_tagged_cl_array(
-                        cla.empty(self.queue.context,
-                                  shape=subary.shape,
-                                  dtype=subary.dtype,
-                                  allocator=self.allocator),
-                        get_cl_axes_from_pt_axes(subary.axes),
-                        subary.tags)
-                else:
-                    key_to_pt_arrays[key] = subary
+                key_to_pt_arrays[key] = subary
 
         # }}}
 


arraycontext/impl/pytato/__init__.py
```diff
@@ -296,17 +296,7 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
...
 
-                if subary.size == 0:
-                    # early exit for 0-sized arrays
-                    key_to_frozen_subary[key] = to_tagged_cl_array(
-                        cla.empty(self.queue.context,
-                                  shape=subary.shape,
-                                  dtype=subary.dtype,
-                                  allocator=self.allocator),
-                        get_cl_axes_from_pt_axes(subary.axes),
-                        subary.tags)
```
inducer, line 307: It seems like this made an effort to copy the axes. How come metadata was still lost?

arraycontext/impl/pytato/__init__.py
```diff
@@ -296,17 +296,7 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
                     raise TypeError(f"{type(self).__name__}.freeze invoked "
                                     f"with non-pytato array of type '{type(array)}'")
 
-                if subary.size == 0:
```
inducer, line 299: How big a win was the early exit?

arraycontext/impl/pytato/__init__.py
```diff
@@ -296,17 +296,7 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
...
-                    key_to_frozen_subary[key] = to_tagged_cl_array(
-                        cla.empty(self.queue.context,
-                                  shape=subary.shape,
-                                  dtype=subary.dtype,
-                                  allocator=self.allocator),
-                        get_cl_axes_from_pt_axes(subary.axes),
-                        subary.tags)
-                else:
-                    key_to_pt_arrays[key] = subary
+                key_to_pt_arrays[key] = subary
```
inducer, line 299: Could you add a comment highlighting what the issue was, to ensure we don't reintroduce it?

inducer, line 299: Would it make sense to add a regression test for this?

inducer (commented):
Thanks! CI is probably not going to fly until https://github.com/inducer/loopy/pull/350/ is in.

arraycontext/impl/pytato/__init__.py
kaushikcfd: Aah yep, the metadata isn't lost. I think this was just a case of misattribution.

arraycontext/impl/pytato/__init__.py
kaushikcfd: This wasn't put in for the sake of performance, but empty domains were interfering with loopy transforms, which is no longer an issue post https://github.com/inducer/pytato/pull/320.

kaushikcfd:
Since this is not solving a real problem, I'll close this. See https://github.com/inducer/arraycontext/pull/167#discussion_r893908967.

inducer:
Sounds good by me! :)

inducer:
Revived in #168.

majosm:
IIRC, the reason this was causing me issues is that by exiting early, it wasn’t bubbling up the axis information from the contained expression to the final result array. Thus any axis tagging done inside was being thrown away.

kaushikcfd:
>  Thus any axis tagging done inside was being thrown away.

@inducer this makes sense. Essentially, we weren't calling `self.transform_dag` to propagate the axes tags.

inducer:
@majosm Thanks for following up! Added a comment to explain this in #169.

PR #169, opened 2022-06-10
Add comment to warn against early exits from freeze

inducer:


arraycontext/impl/pytato/__init__.py
```diff
@@ -296,6 +296,10 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
                     raise TypeError(f"{type(self).__name__}.freeze invoked "
                                     f"with non-pytato array of type '{type(array)}'")
 
+                # Don't be tempted to take shortcuts here for, e.g. empty
```
kaushikcfd, line 299:
```suggestion
                # Don't be tempted to take shortcuts here, for e.g. empty
```

arraycontext/impl/pytato/__init__.py
```diff
@@ -296,6 +296,10 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
...
+                # arrays, as this will inhibit metadata propagation that
+                # may happen in transform_dag, below. See
```
kaushikcfd, line 301:
```suggestion
                # may happen in transform_dag below. See
```
?

kaushikcfd (approved):
Thanks! LGTM.

arraycontext/impl/pytato/__init__.py
inducer: e245f3b7ded208d84eac636fbc1c1b204a94b4ca

arraycontext/impl/pytato/__init__.py
inducer: e245f3b7ded208d84eac636fbc1c1b204a94b4ca

inducer:
Thanks for taking a look!

PR #170, opened 2022-06-10
Use `PrefixNamed` to generate better kernel names in pytato freeze

inducer:
Plus a naming nitpick.

### DIFF
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -309,20 +309,38 @@ class PytatoPyOpenCLArrayContext(_BasePytatoArrayContext):
         try:
             pt_prg = self._freeze_prg_cache[normalized_expr]
         except KeyError:
-            if normalized_expr in self._dag_transform_cache:
-                transformed_dag = self._dag_transform_cache[normalized_expr]
-            else:
+            try:
+                transformed_dag, function_name = \
+                        self._dag_transform_cache[normalized_expr]
+            except KeyError:
                 transformed_dag = self.transform_dag(normalized_expr)
-                self._dag_transform_cache[normalized_expr] = transformed_dag
+
+                from pytato.tags import PrefixNamed
+                name_hint_tags = []
+                for subary in key_to_pt_arrays.values():
+                    name_hint_tags.extend(subary.tags_of_type(PrefixNamed))
+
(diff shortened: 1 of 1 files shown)


inducer:
CI breakage is unrelated, cf. https://github.com/illinois-ceesd/mirgecom/pull/686.

arraycontext/impl/pytato/__init__.py
```diff
@@ -309,20 +309,37 @@ def _record_leaf_ary_in_dict(key: Tuple[Any, ...],
...
-                self._dag_transform_cache[normalized_expr] = transformed_dag
+
+                from arraycontext.metadata import NameHint
+                function_name = None
+                for subary in key_to_pt_arrays.values():
+                    naming_tags = subary.tags_of_type(NameHint)
+                    if naming_tags:
+                        naming_tag, = naming_tags
+                        function_name = f"frozen_{naming_tag.name}"
+                        break
```
kaushikcfd, line 325: I'm not sure I'm quite convinced by this. This logic here is essentially picking some output array's name. I think that will lead to some confusing kernel names, for e.g. freezing a `ConservedVars` could end up having the name `frozen_velocity` and I would argue that that's more confusing than a random name.

kaushikcfd (commented):
Thanks, have some questions.

arraycontext/impl/pytato/__init__.py
inducer: You're right to be skeptical. Would it be more appealing if I made this apply only in the case where there's a single output?

kaushikcfd: > You're right to be skeptical. Would it be more appealing if I made this apply only in the case where there's a single output?

Yep, that would definitely kill the confusion.

For the case of containers, we could also just pick the name of the container class.

inducer: See how you like dbe48e9c627ced5ded7a7a495e86e47c096f6999. It looks for a common prefix and uses that if it exists.

kaushikcfd: That looks OK. Thanks!

kaushikcfd (approved):
LGTM, thanks!

inducer:
Thanks for taking a look!

PR #174, opened 2022-06-17
Filter out unavailable pytest actx factories

alexfikl:
This makes two changes to `pytest_generate_tests_for_array_contexts`

* adds a `PytestArrayContextFactory.is_available` to check if the array context works and filters it out if it doesn't
* does not create a factory per CL device, since that does not quite work for non-CL array contexts (the `jax` actx was supposedly running once for each `pyopencl.Device`)


inducer:
LGTM, thanks!

PR #175, opened 2022-06-19
Added check for ignoring strides of 0 sized arrays

mitkotak:
Strides for zero-sized arrays are ill-defined in numpy
```
import numpy as np
a = np.array([])
a.strides # prints (8,)
```

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -779,7 +779,12 @@ def unflatten(
             # {{{ check strides
 
             if strict and hasattr(template_subary, "strides"):
-                if template_subary.strides != subary.strides:
+                # Checking strides for 0 sized arrays is ill-defined
+                # since they cannot be indexed
+                if (
+                    template_subary.strides != subary.strides
+                    and template_subary.size != 0
+                ):
                     raise ValueError(
                             f"strides do not match template: got {subary.strides}, "
                             f"expected {template_subary.strides}")


arraycontext/container/traversal.py
```diff
@@ -744,7 +744,11 @@ def _unflatten(template_subary: ArrayOrContainerT) -> ArrayOrContainerT:
                     "'template' is too large")
 
             if strict:
-                if template_subary.dtype != ary.dtype:
+                # Checking strides for 0-sized arrays is ill-defined in numpy
+                if (
+                    (template_subary.strides != ary.strides)
+                    and (template_subary.size != 0)
+                ):
```
kaushikcfd, line 751: Not the correct spot to insert this check. Would've been better to directly set the source branch as the branch in https://github.com/kaushikcfd/arraycontext/pull/4 :).

arraycontext/container/traversal.py
```diff
@@ -528,7 +528,6 @@ def freeze(
     :class:`ArrayContainer` *ary*.
 
     :param ary: a :meth:`~ArrayContext.thaw`\ ed :class:`ArrayContainer`.
-
```
kaushikcfd, line 531: Looks like some spurious line insertions/deletions sprung in. Get rid of it?

arraycontext/container/traversal.py
```diff
@@ -779,7 +779,10 @@ def _unflatten(template_subary: ArrayOrContainerT) -> ArrayOrContainerT:
             # {{{ check strides
 
             if strict and hasattr(template_subary, "strides"):
-                if template_subary.strides != subary.strides:
+                if (
+                    (template_subary.strides != subary.strides)
+                    and (template_subary.size != 0)
```
kaushikcfd, line 784: Looks like this patch misses the resolution from https://github.com/kaushikcfd/arraycontext/pull/4#discussion_r896022401. (Also, please explain *why* the strides checking ill-defined.)

arraycontext/container/traversal.py
kaushikcfd: This is fixed in the latest version. Thanks!

arraycontext/container/traversal.py
mitkotak: Yup it got dropped when I was updating my branch. Its resolved now. Thanks !

alexfikl:
Why is this needed now? We fixed a bunch of these issues upstream in `pyopencl` and `loopy` and there's a test that was passing now for everyone at
https://github.com/inducer/arraycontext/blob/d1157d1aed6e9a2c713714c656843c260e7bfa48/test/test_arraycontext.py#L1001

kaushikcfd:
> Why is this needed now? We fixed a bunch of these issues upstream in pyopencl and loopy and there's a test that was passing now for everyone at

Hmm.. I thought checking strides on 0-size arrays didn't make sense as the elements cannot be accessed, so the strides are free to take any value.

As a side note, PyCUDA arrays don't respect NumPy's strides for 0-sized arrays and I didn't quite think PyCUDA was at fault as semantically everything is still correct.

```python
import numpy as np
import pycuda.autoinit  # noqa: F401
import pycuda.gpuarray as cu_np

a = np.zeros((10, 0))
a_gpu = cu_np.to_gpu(a)
print(a.strides)        # prints: (8, 8)
print(a_gpu.strides)    # prints: (0, 8)
```

alexfikl:
> Hmm.. I thought checking strides on 0-size arrays didn't make sense as the elements cannot be accessed, so the strides are free to take any value.

From what I remember, `numpy` ignores strides when looking at contiguity if the array has size 0. For access it seems like it still works with `ary[i, ()]` for a size of `(n, 0)`, but the strides are also pretty much ignored.

```python
print(a.strides)        # prints: (8, 8)
print(a_gpu.strides)    # prints: (0, 8)
```
Yeah, `pyopencl` used to do this too and was changed to match `numpy` a bit better in inducer/pyopencl#514. It still raises an index error on `ary[i, ()]` though.

Either way, I didn't know this was for `pycuda`, so was just curious if something broke!

arraycontext/container/traversal.py
```diff
@@ -779,7 +779,12 @@ def _unflatten(template_subary: ArrayOrContainerT) -> ArrayOrContainerT:
             # {{{ check strides
 
             if strict and hasattr(template_subary, "strides"):
-                if template_subary.strides != subary.strides:
+                # Checking strides for 0 sized arrays is ill-defined
+                # since they cannot be indexed
+                if (
+                    (template_subary.strides != subary.strides)
+                    and (template_subary.size != 0)
```
inducer, line 786:
```suggestion
                    template_subary.strides != subary.strides
                    and template_subary.size != 0
```

inducer:
Thanks for working on this! This change is fine by me, I'd be happy to merge it (pending the small fix above).

Like @alexfikl, I'm a bit curious what brought about the impulse to make this change.

mitkotak:
My CI for PyCUDArrayContext was failing due to the test case mentioned above which led me to investigate 0 sized arrays

> Why is this needed now? We fixed a bunch of these issues upstream in `pyopencl` and `loopy` and there's a test that was passing now for everyone at
> 
> https://github.com/inducer/arraycontext/blob/d1157d1aed6e9a2c713714c656843c260e7bfa48/test/test_arraycontext.py#L1001

inducer:
Thanks!

PR #176, opened 2022-06-19
Make name of doc build flag more global, explain

inducer:


inducer:
Merging since only the doc build is affected and that passed.

PR #177, opened 2022-06-24
Split dataclass_array_container

alexfikl:
This splits the code generation bits out of `dataclass_array_container` so that other packages can more easily implement similar behavior (notably `boxtree` at the moment).

### DIFF
--- a/arraycontext/container/dataclass.py
+++ b/arraycontext/container/dataclass.py
@@ -30,18 +30,23 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Union, get_args
+from typing import Tuple, Union, get_args
 try:
     # NOTE: only available in python >= 3.8
     from typing import get_origin
 except ImportError:
     from typing_extensions import get_origin
 
-from dataclasses import fields
+from dataclasses import Field, is_dataclass, fields
 from arraycontext.container import is_array_container_type
 
 
 # {{{ dataclass containers
+
+def is_array_type(tp: type) -> bool:
+    from arraycontext import Array
+    return tp is Array or is_array_container_type(tp)
+
 
 def dataclass_array_container(cls: type) -> type:
     """A class decorator that makes the class to which it is applied an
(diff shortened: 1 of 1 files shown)


arraycontext/container/dataclass.py
```diff
@@ -51,47 +51,39 @@ def dataclass_array_container(cls: type) -> type:
 
     Attributes that are not array containers are allowed. In order to decide
     whether an attribute is an array container, the declared attribute type
-    is checked by the criteria from :func:`is_array_container_type`.
+    is checked by the criteria from :func:`is_array_container_type`. This
+    includes some support for type annotations:
+
+    * a :class:`typing.Union` of array containers is considered an array container.
+    * other type annotations, e.g. :class:`typing.Optional`, are not considered
+      array containers, even if they wrap one.
```
inducer, line 59: This strikes me as strange. What are the trade-offs that went into this decision? (We should make sure to record them in a comment, once we reach a decision.)

alexfikl: Well, from what I can tell, there are two use cases here which are pretty valid (assuming the type annotations reflect reality):

* If the container is used for arithmetic, it doesn't really make sense to allow `Optional[Container]`, since that would fail in interesting ways. I don't think teaching the arithmetic about `None` is worth it (?).
* If the container is just used to hold data and transfer it to/from `numpy`, like most things in `boxtree` do, then `Optional[Container]` is ok. This would also need some support in `thaw`/`freeze`/`to_numpy`/`from_numpy`, but overall doable.

The reason it ignores `Optional[Container]` right now is because all the array containers we have are used for arithmetic, so it was geared towards that.

Maybe it's worth adding a flag or something like #96?

inducer: I buy that reasoning. Could you add this as a comment?

alexfikl: Added a little note in [b0da8bc](https://github.com/inducer/arraycontext/pull/177/commits/b0da8bc480868171cd574f65933ee72631dfc68e). Let me know what you think!

What about the usage in `boxtree`? Right now (in inducer/boxtree#56), the fields are set to `Container` instead of `Optional[Container]` since they would not be serialized otherwise, but that doesn't seem great. 

We could break `dataclass_array_container` apart a bit to let `boxtree` implement it's own custom version more easily? Basically just add a `inject_container_serialization` function that does the nasty code generation parts.

inducer: I agree. Splitting `dataclass_array_container` seems like a better idea than adding less-than-well-justified implicit behaviors.

alexfikl: Split some of it in [4ba3383](https://github.com/inducer/arraycontext/pull/177/commits/4ba3383f099d564347da51fcad43b4c266eb1c5a).

alexfikl: @inducer Modified the PR to just split parts of `dataclass_array_container` and leave the rest of the functionality the same, i.e. not allow any type annotations. If it can be easily re-implemented in `boxtree`, there's no point in making this version less strict.

inducer:
LGTM, thanks!

PR #180, opened 2022-06-28
Add force eval method.

MTCam:
This change set adds a convenience utility for forcing evaluation of any expressions in an array input. (WIP)

### DIFF
--- a/arraycontext/context.py
+++ b/arraycontext/context.py
@@ -239,6 +239,7 @@ class ArrayContext(ABC):
 
     .. automethod:: freeze
     .. automethod:: thaw
+    .. automethod:: force_evaluate
     .. automethod:: tag
     .. automethod:: tag_axis
     .. automethod:: compile
@@ -462,6 +463,15 @@ class ArrayContext(ABC):
         :return: a function with the same signature as *f*.
         """
         return f
+
+    def force_evaluate(self, array: Array) -> Array:
+        """Force the evaluation of any expression trees in the input *array* and
+        return a 'thaw'ed version of that array.  This function forces evaluation
+        by calling 'freeze' and subsequently 'thaw'.
+
+        See also :func:`arraycontext.freeze`, :func:`arraycontext.thaw`
+        """
+        return self.thaw(self.freeze(array))
 
     # undocumented for now
     @property


arraycontext/context.py
```diff
@@ -463,6 +464,15 @@ def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
         """
         return f
 
+    def force_evaluate(self, array: Array) -> Array:
+        """Force the evaluation of any expression trees in the input *array* and
+        return a 'thaw'ed version of that array.  This function forces evaluation
+        by calling 'freeze' and subsequently 'thaw'.
+
+        See also :func:`arraycontext.freeze`, :func:`arraycontext.thaw`
```
inducer, line 472: Those are deprecated. Refer to the methods instead.

arraycontext/context.py
```diff
@@ -463,6 +464,15 @@ def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
         """
         return f
 
+    def force_evaluate(self, array: Array) -> Array:
```
inducer, line 467: I wonder whether we should call this `freeze_thaw` instead to avoid inventing new words.

arraycontext/context.py
```diff
@@ -463,6 +464,15 @@ def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
...
+        """Force the evaluation of any expression trees in the input *array* and
+        return a 'thaw'ed version of that array.  This function forces evaluation
+        by calling 'freeze' and subsequently 'thaw'.
```
inducer, line 470: Link to those methods.

inducer, line 470: You can talk about evaluation of expression trees as an illustrative example, but not when you define what this does. Think of the documentation triangle: what, how, why. Keep them separate, and in that order.

arraycontext/context.py
inducer, line 467: Thanks for adding a type annotation. Technically, this should now be `ArrayOrContainerT`, and definitely a type variable.  I saw that `freeze` and `thaw` also have incorrect annotations, we should fix those too.

inducer (commented):
Thanks for working on this! A few comments below.

arraycontext/context.py
MTCam: Thanks for having a look @inducer!   I think I've addressed your comments in ([2f142f8](https://github.com/inducer/arraycontext/pull/180/commits/2f142f8b989b8f2c9d47c3a08f787abb5c79863b)).   I tried to annotate with `ArrayOrContainerT`, but there was a circular import issue.

PR #181, opened 2022-06-29
Clean up type annotations, add `freeze_thaw`

MTCam:
Address type annotations comments in #180.

### DIFF
--- a/arraycontext/container/__init__.py
+++ b/arraycontext/container/__init__.py
@@ -3,22 +3,10 @@
 """
 .. currentmodule:: arraycontext
 
-.. class:: ArrayT
-    :canonical: arraycontext.container.ArrayT
-
-    :class:`~typing.TypeVar` for arrays.
-
-.. class:: ContainerT
-    :canonical: arraycontext.container.ContainerT
-
-    :class:`~typing.TypeVar` for array container-like objects.
-
-.. class:: ArrayOrContainerT
-    :canonical: arraycontext.container.ArrayOrContainerT
-
-    :class:`~typing.TypeVar` for arrays or array container-like objects.
-
 .. autoclass:: ArrayContainer
+.. class:: ArrayContainerT
+
+    A type variable with a lower bound of :class:`ArrayContainer`.
 
 .. autoexception:: NotAnArrayContainerError
 
@@ -38,7 +26,22 @@ Context retrieval
 ---------------------------------------------------------
 
 .. autofunction:: register_multivector_as_array_container
+
+.. currentmodule:: arraycontext.container
+
+Canonical locations for type annotations
+----------------------------------------
+
+.. class:: ArrayContainerT
+
+    :canonical: arraycontext.ArrayContainerT
+
+.. class:: ArrayOrContainerT
+
+    :canonical: arraycontext.ArrayOrContainerT
 """
+
+from __future__ import annotations
 
 
 __copyright__ = """
@@ -67,22 +70,24 @@ THE SOFTWARE.
 
 from functools import singledispatch
 from arraycontext.context import ArrayContext
-from typing import Any, Iterable, Tuple, TypeVar, Optional, Union, TYPE_CHECKING
+from typing import Any, Iterable, Tuple, Optional, TypeVar, Protocol, TYPE_CHECKING
 import numpy as np
 
-ArrayT = TypeVar("ArrayT")
-ContainerT = TypeVar("ContainerT")
-ArrayOrContainerT = Union[ArrayT, ContainerT]
+# For use in singledispatch type annotations, because sphinx can't figure out
+# what 'np' is.
+import numpy
+
 
 if TYPE_CHECKING:
     from pymbolic.geometric_algebra import MultiVector
+    from arraycontext import ArrayOrContainer
 
 
 # {{{ ArrayContainer
 
-class ArrayContainer:
-    r"""
-    A generic container for the array type supported by the
+class ArrayContainer(Protocol):
+    """
+    A protocol for generic containers of the array type supported by the
     :class:`ArrayContext`.
 
     The functionality required for the container to operated is supplied via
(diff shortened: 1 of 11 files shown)


arraycontext/context.py
```diff
@@ -134,6 +134,10 @@
     import loopy
 
 
+ArrayT = TypeVar("ArrayT")
```
alexfikl, line 137: There's already an `ArrayT` below, which looks a bit nicer.

arraycontext/context.py
```diff
@@ -308,7 +313,7 @@ def call_loopy(self,
         """
 
     @abstractmethod
-    def freeze(self, array: Array) -> Array:
+    def freeze(self, array: ArrayOrContainerT) -> ArrayOrContainerT:
```
alexfikl, line 316: I'm not sure this is accurate, e.g. `pytato` can take a `pt.Array` and return a `cl.Array` and this constraints the input and output to have the same type.

Maybe something like 
```suggestion
    def freeze(self, array: Union[Array, ContainerT, _ScalarLike]) -> Union[Array, ContainerT, _ScalarLike]:
```
? I'm not sure that works though.. may need to [overload](https://docs.python.org/3/library/typing.html#typing.overload) to get this vaguely accurate..

Typing is hard :(

arraycontext/context.py
```diff
@@ -463,6 +468,22 @@ def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
...
+        arithmetic operations using the execution resources of this context.
+        It does this by calling :meth:`~arraycontext.ArrayContext.freeze` and
+        subsequently :meth:`~arraycontext.ArrayContext.thaw`.
+
+        This method can be used, for example, in the evaluation of any expression
+        trees in the input *array* and producing a 'thaw'\ ed version of the data
+        computed from that evaluation.
+
+        See also :meth:`~arraycontext.ArrayContext.freeze`,
+        :meth:`~arraycontext.ArrayContext.thaw`
```
alexfikl, line 483:
```suggestion
        See also :meth:`freeze` and :meth:`thaw`.
```
? Should work too since they're in the same file.

alexfikl (commented):
Left a couple of comments, but not really sure how to properly type `freeze` and friends :disappointed:

arraycontext/context.py
MTCam: Thanks @alexfikl!   I am not sure what to do with this one.  I think I understand the concern - but not sure how to resolve it properly.  I bent your suggestion just a bit to be consistent with @inducer's comments on #180.

alexfikl: Yeah, I'm not sure it's worth diving too much into this until `mypy` starts complaining or we want to use `--strict`. For now, a "good enough" annotation and some clear docs should do the trick!

inducer: I agree with @alexfikl that typing is hard. And I think arraycontext is sufficiently reliant on Python's dynamic typing that trying to get it fully statically typed is probably a fool's errand. That said, e386416 is my attempt at cleaning things up a bit. I'd be grateful if the both of you could take a look.

arraycontext/container/__init__.py
```diff
@@ -67,21 +52,18 @@
 
 from functools import singledispatch
 from arraycontext.context import ArrayContext
-from typing import Any, Iterable, Tuple, TypeVar, Optional, Union, TYPE_CHECKING
+from typing import Any, Iterable, Tuple, Optional, TypeVar, Protocol, TYPE_CHECKING
```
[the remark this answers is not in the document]
inducer: I'll simply bump the Python compat target to 3.8.

arraycontext/container/__init__.py
```diff
@@ -117,6 +99,14 @@ class ArrayContainer:
         special meaning or behavior.
     """
 
+    # Array containers do not need to have any particular features, but
+    # this type must be distinct from Any, so that the dataclass decorators
+    # can recognize it.
```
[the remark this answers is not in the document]
inducer: See above. This is trying to unify `ArrayContainer` (which we used for dataclass attribute type annotation) and the rest of our annotation infrastructure.

arraycontext/context.py
```diff
@@ -162,35 +183,26 @@ def shape(self) -> Tuple[int, ...]:
...
-    def shape(self) -> Tuple[()]:
+    def size(self) -> int:
         ...
 
     @property
     def dtype(self) -> "np.dtype[Any]":
         ...
 
 
+# deprecated, use ArrayScalar instead
```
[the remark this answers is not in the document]
inducer: 0d316ca3bd9848cc77d1b5f3572c8754f4b2ae4f

arraycontext/context.py
```diff
@@ -162,35 +183,26 @@ def shape(self) -> Tuple[int, ...]:
...
+Scalar = ScalarLike
+
+
+ArrayT = TypeVar("ArrayT", bound=Array)
+ArrayOrContainer = Union[Array, "ArrayContainer"]
+ArrayOrContainerT = TypeVar("ArrayOrContainerT", bound=ArrayOrContainer)
+ArrayOrContainerOrScalar = Union[Array, Any, ScalarLike]
```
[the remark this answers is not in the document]
inducer: 0d316ca3bd9848cc77d1b5f3572c8754f4b2ae4f

arraycontext/container/__init__.py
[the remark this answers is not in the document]
inducer: Done in 16e09bba5aec55177c780b281a564cb325dddf56.

inducer:
Unfortunately, this lights all the subpackages on fire, because they import type aliases from submodules instead of the root. I propose that we make all the types available in the root (as done here) in a separate PR, fix the imports in the subpackages, and then merge this.

inducer:
Ok, #183 is that PR. Empirically, that suffices for grudge. I noticed that all instances of `ArrayOrContainerT` should really have been `ArrayOrContainer`, so I fixed that, too. Thanks for offering to do meshmode and pytential! Once you do them, could you let me know whether you need more symbols in #183?

inducer:
https://github.com/inducer/grudge/pull/270 is the fix for grudge. @MTCam, could you take care of mirgecom?

inducer:
https://github.com/inducer/pytential/pull/163 and https://github.com/inducer/meshmode/pull/341 are in, rerunning CI.

PR #183, opened 2022-06-29
Expose types for subpackages in package root

inducer:
This exposes a few type things in the root so that subpacakges can be fixed, as described here: https://github.com/inducer/arraycontext/pull/181#issuecomment-1170378680.

### DIFF
--- a/arraycontext/__init__.py
+++ b/arraycontext/__init__.py
@@ -29,7 +29,9 @@ THE SOFTWARE.
 """
 
 import sys
-from .context import ArrayContext, Array, Scalar, tag_axes
+from .context import (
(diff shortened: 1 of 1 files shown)


alexfikl:
From a quick grep, meshmode also needs `from .context import ArrayT`

inducer:
Added `Array` and `ArrayT`.

inducer:
(Make sure to change to `Array`, not `ArrayT`, when things aren't required to be a type variable.)

inducer:
Let's call this finished and add more commits later if needed.

inducer:
I've just gone ahead and force-merged this so that we can move on with life. :)

PR #178, opened 2022-06-26
Support containers in to/from_numpy

alexfikl:
This is a bit bigger than it's supposed to be because it moves some things around to avoid a bit of duplicated code, but the main changes are:

* adds support for containers to `actx.to_numpy` and `actx.from_numpy` (deprecating the functions in `traversal.py`)
* adds support for containers to `actx.empty_like` (and `actx.zeros_like`) to avoid confusion with `actx.np.empty_like` (and `actx.np.zeros_like`).
* adds some helpers for `TaggableCLArray`.
* Fix `jax.np.all` and `jax.np.any`: it was just forwarding to `jnp.all` which doesn't work right on containers.

With this, all the functions in `ArrayContext` that should work with containers do.

TODO

* [x] Just forward `actx.empty_like` in `actx.np.empty_like` (same for `zeros_like`)
* [x] Add some tests? Not sure any more are needed since the functions are tested in other places with containers.

PR #184, opened 2022-07-02
Deprecate empty / empty_like / zeros_like

alexfikl:
As discussed in #178, this
* deprecates `empty` and `empty_like`.
* deprecates `actx.np.empty_like` for the contexts that supported that.
* deprecates `actx.zeros_like` in favor of `actx.np.zeros_like`.

### DIFF
--- a/arraycontext/context.py
+++ b/arraycontext/context.py
@@ -299,9 +299,19 @@ class ArrayContext(ABC):
         pass
 
     def empty_like(self, ary: Array) -> Array:
+        from warnings import warn
+        warn(f"{type(self).__name__}.empty_like is deprecated and will stop "
+            "working in 2023. Prefer actx.np.zeros_like instead.",
+            DeprecationWarning, stacklevel=2)
+
         return self.empty(shape=ary.shape, dtype=ary.dtype)
 
     def zeros_like(self, ary: Array) -> Array:
+        from warnings import warn
+        warn(f"{type(self).__name__}.zeros_like is deprecated and will stop "
+            "working in 2023. Use actx.np.zeros_like instead.",
+            DeprecationWarning, stacklevel=2)
+
         return self.zeros(shape=ary.shape, dtype=ary.dtype)
 
     @abstractmethod
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -15,6 +15,39 @@ implementations for:
 :mod:`arraycontext` started life as an array abstraction for use with the
 :mod:`meshmode` unstrucuted discretization package.
 
+Design Guidelines
+-----------------
+
+Here are some of the guidelines we aim to follow in :mod:`arraycontext`. There
+exist numerous other, related efforts, such as the `Python array API standard
+<https://data-apis.org/array-api/latest/purpose_and_scope.html>`__. These
+points may aid in clarifying and differentiating our objectives.
+
+- The array context is about exposing the common subset of operations
+  available in immutable and mutable arrays. As a result, the interface
+  does *not* seek to support interfaces that provide, enable, or are typically
+  used only with in-place mutation.
+
+  For example: The equivalents of :func:`numpy.empty` were deprecated
+  and will eventually be removed.
+
+- Each array context offers a specific subset of of :mod:`numpy` under
+  :attr:`arraycontext.ArrayContext.np`. Functions under this namespace
+  must be unconditionally :mod:`numpy`-compatible, that is, they may not
+  offer an interface beyond what numpy offers. Functions that are
+  incompatible, for example by supporting tag metadata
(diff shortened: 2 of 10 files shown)


arraycontext/context.py
```diff
@@ -299,9 +299,19 @@ def zeros(self,
         pass
 
     def empty_like(self, ary: Array) -> Array:
+        from warnings import warn
+        warn(f"{type(self).__name__}.empty_like is deprecated and will stop "
+            "working in 2023.",
```
alexfikl, line 304: Should this recommend some replacement or something? Nothing came to mind..

inducer: Why not `actx.np.empty_like`?

inducer: Ah! `empty`! Sorry, lost the plot for a second. Maybe suggest `zeros_like`?

alexfikl: Sure, went with that in https://github.com/inducer/arraycontext/compare/b2785c47586f3b0323a5ce27684a2708756cecd2..7283002dee5013b6f1e87c24c2e9335a6b54913c.

inducer (approved):
LGTM, thanks!

inducer:
- @kaushikcfd, do you agree with this? I feel that, implicitly, these changes do some trajectory-setting. I've tried to make that more explicit by writing them down as guidelines in 69a3d944bda37a0f09c587af8e2bd7ced30b2b42. If you feel that this is not a good direction for `arraycontext`, now would be the time to suggest alternatives.
- @alexfikl Same for you. Could you look at the text I've added in 69a3d944bda37a0f09c587af8e2bd7ced30b2b42 and complain if you disagree?

In either case, each of you, please let me know yea or nay so this can get merged.

doc/index.rst
```diff
@@ -15,6 +15,34 @@ implementations for:
...
+  For example: The equivalents of :func:`numpy.empty` were deprecated
+  and will eventually be removed.
+
+- Each array context offers a specific subset of of numpy under
+  :attr:`arraycontext.ArrayContext.np`. Functions under this namespace
+  must be unconditionally :mod:`numpy`-compatible, that is, they may not
+  offer an interface beyond what numpy offers. Functions that are
+  incompatible, for example by supporting tag metadata
+  (cf. :meth:`arraycontext.ArrayContext.einsum`) should live under the
+  :class:`~arraycontext.ArrayContext` directly.
```
kaushikcfd, line 35: Why is this the case? The default argument handling in `actx.einsum` make it compatible with `numpy.einsum`.
(I know there was some discussion about this during early design, but I wasn't heavily involved then. Sorry for the redundancy)

kaushikcfd (approved):
One small query, otherwise this LGTM!

I do like that clearly specifying that using a lot of `if actx.suport_xxx` to branch is a non-pattern. Thanks!

doc/index.rst
inducer: I tried to make this explicit. [einsum](https://documen.tician.de/arraycontext/array_context.html#arraycontext.ArrayContext.einsum) does have a `tagged` kwargs, which, by the letter of this rule, disqualifies it from being under `actx.np` (IMO).

Just trying to explain FWIW, the strictness of numpy compatibility is up for debate. Though IMO, it's a slippery slope if we do allow deviations.

kaushikcfd: I see. Yep, that makes sense. Thanks for explaining.

doc/index.rst
```diff
@@ -15,6 +15,39 @@ implementations for:
...
+  For example: The equivalents of :func:`numpy.empty` were deprecated
+  and will eventually be removed.
+
+- Each array context offers a specific subset of of :mod:`numpy` under
+  :attr:`arraycontext.ArrayContext.np`. Functions under this namespace
+  must be unconditionally :mod:`numpy`-compatible, that is, they may not
+  offer an interface beyond what numpy offers. Functions that are
+  incompatible, for example by supporting tag metadata
+  (cf. :meth:`arraycontext.ArrayContext.einsum`) should live under the
+  :class:`~arraycontext.ArrayContext` directly.
```
alexfikl, line 40: It doesn't seem like these will grow too much, but should they get their own namespace too? Something like `actx.ac.einsum`?

I'm mostly thinking how `jax` also has `jax.lax` for more low-level specific stuff.

alexfikl (commented):
Added a little (non-blocking) question, but overall that looks sensible and we've been following most of it already!

doc/index.rst
inducer: I don't see a big reason at this point, mostly because I'm not sure I see a big distinction between numpy-likes that we've extended and our own interfaces. And I agree, I'm not sure we'll have tons of them, so the likelihood of clashes is low.

inducer:
Thanks to both of you for taking a look!

PR #189, opened 2022-08-31
PytatoPyOpenCLArrayContext: use SVM allocator if available, limit arg size for GPUs

matthiasdiener:
Needs:

- ~~https://github.com/inducer/pytato/pull/359~~ (maybe not needed)
- ~~https://github.com/inducer/loopy/pull/642~~

### DIFF
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -417,9 +417,9 @@ class LazilyPyOpenCLCompilingFunctionCaller(BaseLazilyCompilingFunctionCaller):
         with ProcessLogger(logger, f"generate_loopy for '{prg_id}'"):
             import pyopencl as cl
             dev = self.actx.context.devices[0]
-            limit = dev.max_parameter_size
             target = None
             if dev.type & cl.device_type.GPU:
+                limit = dev.max_parameter_size
                 # Leave some extra space since our sizes are estimates
                 target = lp.PyOpenCLTarget(limit_arg_size_nbytes=limit//2)
 
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -418,7 +418,8 @@ class LazilyPyOpenCLCompilingFunctionCaller(BaseLazilyCompilingFunctionCaller):
             import pyopencl as cl
             dev = self.actx.context.devices[0]
             target = None
-            if dev.type & cl.device_type.GPU:
+            if (dev.type & cl.device_type.GPU
+                    and cl.characterize.has_coarse_grain_buffer_svm(dev)):
                 limit = dev.max_parameter_size
                 # Leave some extra space since our sizes are estimates
                 target = lp.PyOpenCLTarget(limit_arg_size_nbytes=limit//2)
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -415,22 +415,13 @@ class LazilyPyOpenCLCompilingFunctionCaller(BaseLazilyCompilingFunctionCaller):
                 prg_id, "pre_generate_loopy", pt_dict_of_named_arrays)
 
         with ProcessLogger(logger, f"generate_loopy for '{prg_id}'"):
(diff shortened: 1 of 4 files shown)


arraycontext/impl/pytato/compile.py
```diff
@@ -415,12 +415,23 @@ def _dag_to_transformed_pytato_prg(self, dict_of_named_arrays, *, prg_id=None):
                 prg_id, "pre_generate_loopy", pt_dict_of_named_arrays)
 
         with ProcessLogger(logger, f"generate_loopy for '{prg_id}'"):
+            import pyopencl as cl
+            dev = self.actx.context.devices[0]
```
inducer, line 419: Safer to get it from the queue, which only has one device.

arraycontext/impl/pytato/compile.py
```diff
@@ -415,12 +415,23 @@ def _dag_to_transformed_pytato_prg(self, dict_of_named_arrays, *, prg_id=None):
...
+                # Leave some extra space since our sizes are estimates
+                target = lp.PyOpenCLTarget(limit_arg_size_nbytes=limit//2)
+
             pytato_program = pt.generate_loopy(
                     pt_dict_of_named_arrays,
                     options=lp.Options(
                         return_dict=True,
                         no_numpy=True),
-                    function_name=_prg_id_to_kernel_name(prg_id))
+                    function_name=_prg_id_to_kernel_name(prg_id),
```
inducer, line 432: Factor the target construction into a memoized function and also use that in `freeze`?

inducer (commented):
This looks good generally. Two style nits below.

Two questions:
- Does it work?
- How come it's marked as draft?

arraycontext/impl/pytato/compile.py
```diff
@@ -415,12 +415,23 @@ def _dag_to_transformed_pytato_prg(self, dict_of_named_arrays, *, prg_id=None):
...
 
         with ProcessLogger(logger, f"generate_loopy for '{prg_id}'"):
+            import pyopencl as cl
+            dev = self.actx.context.devices[0]
+            target = None
+            if (dev.type & cl.device_type.GPU
+                    and cl.characterize.has_coarse_grain_buffer_svm(dev)):
+                limit = dev.max_parameter_size
+                # Leave some extra space since our sizes are estimates
+                target = lp.PyOpenCLTarget(limit_arg_size_nbytes=limit//2)
```
inducer, line 425: Hmm, upon second thought: We can only pass this if we're sure that the memory allocated is actually SVM. So this has to get involved with memory pool creation.

arraycontext/impl/pytato/compile.py
matthiasdiener: Done in [1e54ce4](https://github.com/inducer/arraycontext/pull/189/commits/1e54ce49945684008ab48a1a7f2793fc5590a702)

arraycontext/impl/pytato/compile.py
matthiasdiener: Done in Done in [1e54ce4](https://github.com/inducer/arraycontext/pull/189/commits/1e54ce49945684008ab48a1a7f2793fc5590a702)

arraycontext/impl/pytato/compile.py
matthiasdiener: Should we do this here or as part of https://github.com/inducer/loopy/pull/642 ?

matthiasdiener:
> * Does it work?

As far as I can tell, yes.

> * How come it's marked as draft?

These still need the other PRs to be merged first I think.

arraycontext/impl/pytato/compile.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/189/commits/faba326cd2142d4072c1132c523e6c82553621df and https://github.com/inducer/arraycontext/pull/189/commits/61038f97575019bc95903995f778ff1283a2c014

matthiasdiener:
This is ready for another review @inducer

inducer:
Thanks!

PR #191, opened 2022-09-10
Rename AllocatorInterface to AllocatorBase

alexfikl:
Should fix the docs
https://github.com/inducer/boxtree/runs/8283995087?check_suite_focus=true


inducer:
Thx!

PR #194, opened 2022-09-20
Make pytato a soft dependency again

alexfikl:
Introduced in #189.

Better to move this to `impl/pytato/utils.py`? Seems like a better place for it.


inducer:
Good point, sorry about that! Also, completely agreed about `utils`. See #195, closing this.

PR #198, opened 2022-09-23
PytatoPyOpenCLArrayContext: don't trust the arg limit reported by the GPU

matthiasdiener:
Apparently, CUDA doesn't like it when argument sizes get close to the reported limit.

Avoids PTX/JIT errors of the type:
CUDA_ERROR_INVALID_IMAGE: device kernel image is invalid
CUDA_ERROR_FILE_NOT_FOUND: file not found

Partially addresses https://github.com/illinois-ceesd/mirgecom/issues/679

### DIFF
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -388,15 +388,19 @@ class PytatoPyOpenCLArrayContext(_BasePytatoArrayContext):
                     and cl_char.has_coarse_grain_buffer_svm(dev))):
 
             if dev.max_parameter_size == 4352:
-                # Nvidia devices and PTXAS advertise a limit of 4352 bytes,
+                # Nvidia devices and PTXAS declare a limit of 4352 bytes,
                 # which is incorrect. The CUDA documentation at
                 # https://docs.nvidia.com/cuda/cuda-c-programming-guide/#function-parameters
-                # advertises a limit of 4KB, which is also incorrect.
+                # mentions a limit of 4KB, which is also incorrect.
                 # As far as I can tell, the actual limit is around 4080
                 # bytes, at least on a K40. Reducing the limit further
-                # in order ot be on the safe side.
+                # in order to be on the safe side.
 
                 limit = 4096-200
+
+                from warnings import warn
+                warn("Running on an Nvidia GPU, reducing the argument "
+                    f"size limit from 4352 to {limit}.")
             else:
                 limit = dev.max_parameter_size
 
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -396,6 +396,10 @@ class PytatoPyOpenCLArrayContext(_BasePytatoArrayContext):
                 # bytes, at least on a K40. Reducing the limit further
                 # in order to be on the safe side.
 
+                # Note that the naming convention isn't super consistent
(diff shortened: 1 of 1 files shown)


inducer:
Interesting. That's a weird error message for "too many arguments".

inducer:
LMK when you think this is ready.

matthiasdiener:
I have tested this and it seems to return us to the previous state described in https://github.com/illinois-ceesd/mirgecom/issues/679. This was a regression from my gist which set a fixed limit of 1024 on all devices. @MTCam also experimented with this PR. This is ready for review.

arraycontext/impl/pytato/__init__.py
```diff
@@ -387,7 +387,19 @@ def get_target(self):
                     self.using_svm and dev.type & cl.device_type.GPU
                     and cl_char.has_coarse_grain_buffer_svm(dev))):
 
-            limit = dev.max_parameter_size
+            if dev.max_parameter_size == 4352:
```
inducer, line 390: Add a check that there's an "Nvidia" anywhere in the platform or device name.

arraycontext/impl/pytato/__init__.py
```diff
@@ -387,7 +387,19 @@ def get_target(self):
...
+                # Nvidia devices and PTXAS advertise a limit of 4352 bytes,
+                # which is incorrect. The CUDA documentation at
+                # https://docs.nvidia.com/cuda/cuda-c-programming-guide/#function-parameters
+                # advertises a limit of 4KB, which is also incorrect.
+                # As far as I can tell, the actual limit is around 4080
+                # bytes, at least on a K40. Reducing the limit further
+                # in order ot be on the safe side.
+
+                limit = 4096-200
```
inducer, line 399: Emit a warning that this is happening.

inducer (approved):
Thanks. LGTM with two minor tweaks.

arraycontext/impl/pytato/__init__.py
matthiasdiener: I think this won't work easily since naming across pocl/Nvidia CL is a bit wonky, e.g.

```
<pyopencl.Device 'NVIDIA GeForce GTX TITAN X' on 'Portable Computing Language' at 0x55986cb63fb0>
```

but

```
<pyopencl.Device 'Tesla K40c' on 'Portable Computing Language' at 0x55cd4e26cc20>
```

(ie, no "Nvidia" in the names on the K40)

We could add 'Tesla' to the check, and hope for the best, but imho, another device having 4352 as the limit is really unlikely, so maybe it is enough.

inducer: Ok, fair. But do add a comment about your difficulty adding a naming-based criterion.

arraycontext/impl/pytato/__init__.py
matthiasdiener: Added in [1b94232](https://github.com/inducer/arraycontext/pull/198/commits/1b94232331f20911510a4b17a72e0f5579073c9a) (we could also make this a `logger.info` message)

arraycontext/impl/pytato/__init__.py
matthiasdiener: Added in https://github.com/inducer/arraycontext/pull/198/commits/84587db2c4f65ea7875e8d56c98f6a561fbce9ed

inducer:
LGTM, thanks!

PR #199, opened 2022-09-26
Pin mypy to git for issue with Py3.10.7

inducer:
see https://github.com/python/mypy/issues/13627


inducer:
No longer needed as mypy 0.981 is out.

Issue #196, opened 2022-09-21
Regression: "Out of host memory" on Nvidia ICD


inducer:
- 1593a67b7f2a97e26d4f88fb70bf711e7682a395 (just before the #189) is known good
- 167ae7c6ee7c96118ed7308e3b9e74a8d199088f is broken

Here's a log of the CI failure: https://gitlab.tiker.net/inducer/arraycontext/-/jobs/462367

Host memory grows to 6-ish gigabytes when running this by hand, and then process keels over. If run with `--sw`, this fails as before, but it succeeds upon resuming when rerun. I suspect there's a memory leak here somewhere.

@matthiasdiener Could you take a look?

matthiasdiener:
It looks like Nvidia CL on `bock` does not like creating multiple CL contexts with SVM allocations. The following program quickly fills up the GPU on Nvidia CL, but not with pocl-cuda or pocl:

```python 
import pyopencl as cl
import time
from pyopencl.tools import SVMAllocator, ImmediateAllocator


while True:
    ctx = cl.create_some_context() # putting this before the loop is fine
    queue = cl.CommandQueue(ctx)
    allocator = SVMAllocator(queue.context, queue=queue)
    # allocator = ImmediateAllocator(queue)  # using normal allocations is fine
    x = allocator(2*1024*1024*1024)
    # x.release()
    x.enqueue_release()
    queue.finish()
    del queue
    del ctx

    time.sleep(2)
```

inducer:
Thanks for digging into this! That's helpful as an observation. I'll look at a CL call trace to see if that looks OK, and if so... not sure that we can do much?

inducer:
I'm not sure I like this solution, but I don't see an alternative: Could you do a PR to turn off use of SVM on Nvidia CL, with a pointer to this issue?

I've put together https://gitlab.tiker.net/inducer/arraycontext/-/merge_requests/6 to make sure we maintain at least some coverage of the SVM code paths.

PR #200, opened 2022-09-27
disable SVM on Nvidia CL when running pytest

matthiasdiener:
Partially addresses #196.

### DIFF
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -260,26 +260,37 @@ class PytatoPyOpenCLArrayContext(_BasePytatoArrayContext):
         self.using_svm = None
 
         if allocator is None:
+            import sys
+
             from pyopencl.characterize import has_coarse_grain_buffer_svm
             has_svm = has_coarse_grain_buffer_svm(queue.device)
-            if has_svm:
-                self.using_svm = True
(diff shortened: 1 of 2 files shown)


inducer:
LGTM, thanks!

arraycontext/impl/pytato/__init__.py
```diff
@@ -260,26 +260,37 @@ def __init__(
...
 
         if allocator is None:
+            import sys
+
             from pyopencl.characterize import has_coarse_grain_buffer_svm
             has_svm = has_coarse_grain_buffer_svm(queue.device)
-            if has_svm:
-                self.using_svm = True
+            is_nvidia_cl_with_pytest = (queue.device.platform.name == "NVIDIA CUDA"
+                                        and "pytest" in sys.modules)
```
alexfikl, line 268: Shouldn't this logic be in one of the `PytestArrayContextFactory`? It feels weird to put it in here.

inducer: I agree that that's probably better.

matthiasdiener: What do you think of https://github.com/inducer/arraycontext/pull/200/commits/618f16a20c8aad0a8a89f3a306f8aba05fb6d8ee ?

alexfikl: Much nicer!

PR #201, opened 2022-09-27
pytato.compile: do not assume that the callable will be named

kaushikcfd:
See <https://github.com/inducer/grudge/issues/287>.


inducer:
Thx!

PR #203, opened 2022-10-14
Ensure ArrayContext is not hashable

alexfikl:
Custom objects have some hash based on their `id()`
https://docs.python.org/3/glossary.html#term-hashable

x-ref: inducer/meshmode#354
cc: @kaushikcfd


inducer:
Thanks!

PR #206, opened 2022-10-31
dataclasses: Use default_factory instead of default

kaushikcfd:
Running into errors like:
```
ValueError: mutable default <class 'numpy.ndarray'> for field y is not allowed: use default_factory
```


kaushikcfd:
This is needed because of a breaking backwards compatibility in [Python-3.11](https://docs.python.org/3/whatsnew/3.11.html#dataclasses).

inducer:
Thanks!

PR #208, opened 2022-11-13
Typing fixes for mypy 0.990

inducer:


arraycontext/container/traversal.py
```diff
@@ -682,7 +682,7 @@ def _flatten(subary: ArrayOrContainer) -> List[Array]:
                 # since it is the one responsible for the actual memory layout
                 if hasattr(subary_c, "strides"):
                     # Mypy has a point: nobody promised a strides attr.
```
kaushikcfd, line 684:
```suggestion
```

inducer: Done, thanks!

PR #210, opened 2022-12-13
Initial stab at actx.trace_call

MTCam:
Augments inducer/pytato/#364  by adding array context version of `trace_call` that handles array container types.

### DIFF
--- a/arraycontext/context.py
+++ b/arraycontext/context.py
@@ -520,6 +520,12 @@ class ArrayContext(ABC):
         """
         return f
 
+    # Supporting interface for function/call tracing in actx implementations
+    def trace_call(self, f: Callable[..., Any],
+                   *args, identifier=None, **kwargs):
+        """Returns the result of the called function *f* with the specified args."""
+        return f(*args, **kwargs)
+
     # undocumented for now
     @property
     @abstractmethod
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -44,28 +44,104 @@ THE SOFTWARE.
 
 import abc
 import sys
-from typing import (Any, Callable, Union, Tuple, Type, FrozenSet, Dict, Optional,
-                    TYPE_CHECKING)
-
+from typing import (  # noqa
+    Any, Callable, Dict, FrozenSet, Tuple, Type, Union, TypeVar, Optional,
+    Hashable, Sequence, ClassVar, Iterator, Iterable, Mapping,
+    TYPE_CHECKING
+)
 import numpy as np
 from pytools.tag import ToTagSetConvertible, normalize_tags, Tag
 
-from arraycontext.context import ArrayContext, Array, ArrayOrContainer, ScalarLike
-from arraycontext.container.traversal import (rec_map_array_container,
-                                              with_array_context)
+from arraycontext.context import (
+    ArrayT, ArrayContext, Array, ArrayOrContainer, ScalarLike
+)
+from arraycontext.container.traversal import (
+    rec_map_array_container,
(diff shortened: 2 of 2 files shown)


kaushikcfd:
Is this something that is still being worked on? If yes, I can help with some stuff here, I have some bandwidth for it this week.

If not, then what's the plan for addressing https://github.com/illinois-ceesd/mirgecom/issues/783? (I don't think compilation times of those orders should be tolerated)

MTCam:
> Is this something that is still being worked on? If yes, I can help with some stuff here, I have some bandwidth for it this week.
> 
> If not, then what's the plan for addressing [illinois-ceesd/mirgecom#777](https://github.com/illinois-ceesd/mirgecom/issues/777)? (I don't think compilation times of those orders should be tolerated)

I pushed and updated my latest on this.  My apologies for the state of it; it is very much wip.

I had a fork/branch of meshmode with a simple test of using the actx-provided `trace_call`.   
https://github.com/MTCam/meshmode/tree/use-actx-trace-call

kaushikcfd:
> I pushed and updated my latest on this. My apologies for the state of it; it is very much wip.

Thanks! If you want, I can take over and maybe it will be faster that way? My ETA for this is the weekend. However, if you think you want to get accustomed to these parts of the codebase, I'm happy to step away.

MTCam:
> If not, then what's the plan for addressing [illinois-ceesd/mirgecom#777](https://github.com/illinois-ceesd/mirgecom/issues/777)? (I don't think compilation times of those orders should be tolerated)

The compilation times seem well under control for our current prediction I blv.  We need to run some tests at scale to make sure that this is the case, but currently the build times are hovering around less than an hour for high order 3D with all physics and features enabled on 128 ranks.  That is well into the prediction-enabling realm, and also not a giant bottleneck.   The build time for serial is around 600-1000s, and 3000+ for many ranks.

I understand the current situation is brought about mostly by mitigation and not a permanent solution.   1-dimensional domain decomposition is currently limiting our max number of MPI neighbors to 2, minimizing the impact of the splatted flux dags for partition boundaries.   It does seem like, for our problem and geometry, we could stick with that strategy indefinitely.   

Solving dag splat I think will help us avoid splatting the flux calculation for each unique domain boundary.  That would be useful.

MTCam:
> > I pushed and updated my latest on this. My apologies for the state of it; it is very much wip.
> 
> Thanks! If you want, I can take over and maybe it will be faster that way? My ETA for this is the weekend. However, if you think you want to get accustomed to these parts of the codebase, I'm happy to step away.

I would be delighted for you to take over and am very interested in following what happens here.  I would get back to this after the review, but won't be anywhere near as efficient at getting it going.  I believe it is close to functioning as a demo - but not close to final form.

kaushikcfd:
> but currently the build times are hovering around less than an hour for high order 3D with all physics and features enabled on 128 ranks

Compilation times of 1 hour (even 20 minutes tbh) are very bad and something that must be fixed ASAP. Another problem that is hovering around is that this also leads to a poorly generated device code, as each sub-array from a neighboring rank is assigned a small-ish buffer, and launching GPU kernels for such small problem sizes brings down the throughput.

> I would be delighted for you to take over and am very interested in following what happens here. 

Yep, let's do this then.

kaushikcfd:
See #221.

kaushikcfd:
In #221, I added a test under `examples/`, which demonstrates the intended purpose to:

```console
(py311_env) [line@line examples]$ python how_to_outline.py
[Pre-concatenation] Number of nodes = 32701
[Post-concatenation] Number of nodes = 3710
```

I think we can close this PR.

PR #211, opened 2022-12-14
jax: change DeviceArray to jnp.ndarray

matthiasdiener:
See e.g. https://github.com/inducer/pytato/actions/runs/3695907762/jobs/6258890619

Background: https://jax.readthedocs.io/en/latest/jax_array_migration.html

Needs: 
- https://github.com/inducer/pytato/pull/405

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -13,4 +13,4 @@ dependencies:
 - pyopencl
 - islpy
 - pip
-- jax <0.4
+- jax
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -13,4 +13,4 @@ dependencies:
 - pyopencl
 - islpy
 - pip
-- jax
+- jax <0.4
--- a/arraycontext/impl/jax/__init__.py
+++ b/arraycontext/impl/jax/__init__.py
@@ -54,8 +54,8 @@ class EagerJAXArrayContext(ArrayContext):
     def __init__(self) -> None:
         super().__init__()
 
-        from jax import Array
-        self.array_types = (Array, )
+        from jax import Array as JAXArray
+        self.array_types = (JAXArray, )
 
     def _get_fake_numpy_namespace(self):
         from .fake_numpy import EagerJAXFakeNumpyNamespace
--- a/arraycontext/impl/jax/__init__.py
+++ b/arraycontext/impl/jax/__init__.py
@@ -54,8 +54,8 @@ class EagerJAXArrayContext(ArrayContext):
     def __init__(self) -> None:
         super().__init__()
 
-        from jax.numpy import DeviceArray
-        self.array_types = (DeviceArray, )
+        from jax import Array
+        self.array_types = (Array, )
 
     def _get_fake_numpy_namespace(self):
         from .fake_numpy import EagerJAXFakeNumpyNamespace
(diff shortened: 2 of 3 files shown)


arraycontext/impl/jax/__init__.py
```diff
@@ -54,8 +54,8 @@ class and performs all array operations eagerly. See
     def __init__(self) -> None:
         super().__init__()
 
-        from jax.numpy import DeviceArray
-        self.array_types = (DeviceArray, )
+        from jax import Array
+        self.array_types = (Array, )
```
alexfikl, line 58: Can this use `jnp.ndarray`? From what I know that aliases to the correct underlying class on all versions.

.test-conda-env-py3.yml
```diff
@@ -13,4 +13,4 @@ dependencies:
 - pyopencl
 - islpy
 - pip
-- jax
```
inducer, line 16: Do we still need this? I suspect we want to be on the "new" side of their [`Array` migration](https://jax.readthedocs.io/en/latest/jax_array_migration.html), and have version constraints to match.

matthiasdiener: I kept the version restriction on jax in this PR for the moment to see if jax 0.3 can handle the small code changes in this  PR as well. If so, we might not need a version constraint at all.

arraycontext/impl/jax/__init__.py
matthiasdiener: Both `jax.numpy.ndarray` and `jax.Array` seem to work. I'm not sure which one is better.

alexfikl: `jnp.ndarray` should alias to `DeviceArray` on older versions and `Array` on newer versions of jax. I guess it depends on what versions we want to support?

arraycontext/impl/jax/__init__.py
```diff
@@ -54,8 +54,8 @@ class and performs all array operations eagerly. See
     def __init__(self) -> None:
         super().__init__()
 
-        from jax.numpy import DeviceArray
-        self.array_types = (DeviceArray, )
+        from jax import Array as JAXArray
```
alexfikl, line 57: Looking at the changelog (https://jax.readthedocs.io/en/latest/changelog.html), it seems like `jax.Array` was introduced in 0.3.18. 

Since this imports `jax.Array` unconditionally, maybe we should add a version constraint? i.e. `jax>0.3.18` or something?

matthiasdiener: As far as I can see, the only installation-related place where `jax` is mentioned is in `.test-conda-env-py3.yml`. We could add the constraint there. On the other hand, if we used `jnp.ndarray`, would we need a constraint at all?

arraycontext/impl/jax/__init__.py
matthiasdiener: That's a good point, thanks! I changed it to `jnp.ndarray` in [45482ab](https://github.com/inducer/arraycontext/pull/211/commits/45482ab962934c5a6755412ea45b397c395fcdb8).

arraycontext/impl/jax/__init__.py
alexfikl: Yeah, with `jnp.ndarray` it should be fine throughout :fingers-crossed:

inducer:
Thanks!

PR #212, opened 2022-12-25
Fix swapped `map`/`reduce_func` in `test_container_scalar_map`

inducer:
This started (rightfully) failing with numpy 1.24. I'm not even sure what the prior code was trying to accomplish. I'm still not sure I get what it's trying to do now.

cc @alexfikl as the person who last touched that code.


inducer:
Merging right away because CI on main is broken by this.

alexfikl:
Hm, I think that just randomly tested that the traversal doesn't crash anywhere for weirder inputs. I'll take a look and see how to modify them to actually check the result is correct!

PR #213, opened 2022-12-25
Improve scalar map_reduce_array_container

alexfikl:
Hopefully made this a bit more understandable now: it now checks some total size of the array container.

Followup to #212.

### DIFF
--- a/test/test_arraycontext.py
+++ b/test/test_arraycontext.py
@@ -767,9 +767,10 @@ def test_array_context_einsum_array_tripleprod(actx_factory, spec):
 # {{{ array container classes for test
 
 
-def test_container_scalar_map(actx_factory):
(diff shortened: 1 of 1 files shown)


test/test_arraycontext.py
```diff
@@ -778,16 +779,16 @@ def test_container_scalar_map(actx_factory):
             map_reduce_array_container, rec_map_reduce_array_container,
             )
 
-    for ary in arys:
+    for size, ary in zip(expected_sizes, arys[:-1]):
```
alexfikl, line 782: The last of `arys` is just a scalar and `actx.np.array_equal` doesn't work on scalars, so this skips it now.

inducer:
Thx!

PR #214, opened 2023-01-11
Add a dumb container stringifier

alexfikl:
Used this to debug which fields were not getting recognized as "containers" in the `boxtree.Tree` (which has a lot of fields). For the test case in there, it shows
```
root (SomeOtherContainer)
 +-- disk (SomeContainer)
 |   +-- points (ndarray)
 +-- circle (SomeContainer)
     +-- points (ndarray)
```

Not convinced it's generally useful at all, so feel free to close this!

### DIFF
--- a/arraycontext/__init__.py
+++ b/arraycontext/__init__.py
@@ -44,8 +44,8 @@ from .container.traversal import (
     multimap_array_container, multimap_reduce_array_container,
     multimapped_over_array_containers, outer, rec_map_array_container,
     rec_map_reduce_array_container, rec_multimap_array_container,
-    rec_multimap_reduce_array_container, thaw, to_numpy, unflatten,
-    with_array_context)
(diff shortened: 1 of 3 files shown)


alexfikl:
> I actually like the idea enough to be willing to maintain this.

Great! I'll try to remove the dependency on that `asciitree` package, since it should be fairly easy to just print the thing, and ping you then.

alexfikl:
@inducer Well, that turned out to just work! It now prints
```
root (SomeOtherContainer)
 +-- disk (SomeContainer)
 |   +-- points (ndarray)
 |   +-- centers (ArrayWrapper)
 |       +-- ary (ndarray)
 +-- circle (SomeContainer)
 |   +-- points (ndarray)
 |   +-- centers (ArrayWrapper)
 |       +-- ary (ndarray)
```
with some extra `|` at the end there compared to `asciitree`, but not too bothered.

inducer:
Thx!

PR #215, opened 2023-01-11
Enable and apply isort

alexfikl:
The `isort` commit is very noisy. Worth it?


inducer:
Thanks for doing this. I'd say it's worth it. It's less noisy than I'd expected, and we don't have a lot of big outstanding PRs at the moment.

PR #217, opened 2023-01-22
Implement Batched Einsum ArrayContext

kaushikcfd:
This is a draft because:
- [ ] Includes commits from #216


arraycontext/impl/pytato/batched_einsum/utils.py
```diff
@@ -0,0 +1,647 @@
...
+                                database: str) -> lp.LoopKernel:
+    import feinsum as fnsm
+
+    from arraycontext.impl.pytato.split_actx.utils import (
+        _get_iname_pos_from_loop_nest)
+
+    domain = t_unit[kernel_name].get_inames_domain(loop_nest.inames)
+    if frozenset(domain.get_var_dict()) != loop_nest.inames:
+        raise NotImplementedError("Current implementation requires"
+                                  " ")
```
inducer, line 512: Requires what?

arraycontext/impl/pytato/batched_einsum/utils.py
```diff
@@ -0,0 +1,647 @@
...
+    iname_pos = _get_iname_pos_from_loop_nest(t_unit[kernel_name], loop_nest)
+    subscripts = ", ".join(sorted(loop_nest.inames, key=lambda x: iname_pos[x]))
+    subst_rule_args = ", ".join(f"_{i}" for i in range(len(loop_nest.inames)))
+
+    template_kernel = lp.make_kernel(
+        [domain],
+        f"""
+        in0({subst_rule_args}) := 3.141592653589793
+        in1({subst_rule_args}) := 2.718281828459045
```
inducer, line 521: ?

arraycontext/impl/pytato/batched_einsum/utils.py
```diff
@@ -0,0 +1,647 @@
...
+            idx_tags: Dict[fnsm.EinsumAxisAccess, Tag] = {}
+            for acc_descr, name_in_einsum in einsum.index_names.items():
+                lpy_iname = subst_map[name_in_einsum]
+                lpy_iname_tag, = kernel.iname_tags_of_type(lpy_iname, tag_t)
+                idx_tags[acc_descr] = lpy_iname_tag
+
+            tagged_einsum = EinsumWithAxesTagged(einsum, Map(idx_tags))
+
+            for acc_descr, name_in_einsum in einsum.index_names.items():
+                bucket = (tagged_einsum, acc_descr)
```
inducer, line 235: I am trying to understand the logic here. This appears to be behind issues that @majosm is encountering when trialing this array context in mirgecom, it appears to want to fuse loops of different lengths. And if there are multiple matches to an einsum template within a Loopy kernel, it seems that this will happily toss all inames in analogous "access roles" into the same bucket, irrespective of whether they have the same length.

Beyond the obvious correctness concern, I also wonder about the thought process behind "toss everybody into one fusible bucket"... there might be many many matches within a big compile. All those loops would get fused, resulting in a kernel that may access many different variables, resulting in large amounts of state space use. I didn't see logic to pry these loops back apart if the batch-of-einsums gets too big. What's the intention here? I would be grateful if you could shed some light. Thanks!

kaushikcfd: > it appears to want to fuse loops of different lengths

It cannot fuse loops with different lengths since the `tagged_einsum` would be different for those inames.

> And if there are multiple matches to an einsum template within a Loopy kernel,

We map an iname to a `(normalized_einsum, acc_descr_of_the_iname_in_the_normalized_einsum)`, this tuple should be unique to an einsum. 


> I didn't see logic to pry these loops back apart if the batch-of-einsums gets too big.

We would rely on a tuned `feinsum` transformation to determine of the suitable loop-distribution factor.

arraycontext/impl/pytato/batched_einsum/utils.py
kaushikcfd: Values for $\pi$ and $e$ upto double-precision :). No good?

arraycontext/impl/pytato/batched_einsum/utils.py
inducer: > It cannot fuse loops with different lengths since the `tagged_einsum` would be different for those inames.

Could it be that the "template einsum"  is using a parametric size, and that's not being specialized for matching purposes? It's definitely trying to merge loops of different length in a particular situation for which @majosm unearthed a reproducer (https://github.com/illinois-ceesd/mirgecom/issues/967). I'll do some more digging.

> We would rely on a tuned feinsum transformation to determine of the suitable loop-distribution factor.

That's a sensible approach.

kaushikcfd: > Could it be that the "template einsum" is using a parametric size, and that's not being specialized for matching purposes?

Yep. That's exactly what happened over here. (Thanks!). Loop Fusion should also have been less dumb, see: https://gist.github.com/kaushikcfd/c2707106c47f1fd19d70dfcfd77be845. I'll fix both these things.

kaushikcfd: See the failing reproducer in [e4cf24b](https://github.com/inducer/arraycontext/pull/217/commits/e4cf24b8c1f64f1734cc223d55531d59024bf515) and the associated fix in caaa471.

Thanks for the report!

inducer: Didn't mean to make you do work! But nonetheless thanks for looking into it, and for coming up with a fix. :slightly_smiling_face:

PR #218, opened 2023-02-02
Register real/imag as pytato functions

kaushikcfd:


arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -53,7 +53,7 @@ class PytatoFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
     _pt_funcs = frozenset({
         "sin", "cos", "tan", "arcsin", "arccos", "arctan",
         "sinh", "cosh", "tanh", "exp", "log", "log10",
-        "sqrt", "abs", "isnan"
+        "sqrt", "abs", "isnan", "real", "imag", "conj",
```
inducer, line 56: `real` will return a real `dtype`. Will this use the correct result dtype?

kaushikcfd: Yep, `pt.real`, `pt.imag` are numpy-dtype compatible.

PR #219, opened 2023-02-19
Add logical_(or|and|not) to (Pytato|PyOpenCL)ArrayContext

kaushikcfd:
- [x] Needs https://github.com/inducer/pyopencl/pull/679

### DIFF
--- a/arraycontext/impl/pyopencl/fake_numpy.py
+++ b/arraycontext/impl/pyopencl/fake_numpy.py
@@ -259,6 +259,15 @@ class PyOpenCLFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
     def not_equal(self, x, y):
         return rec_multimap_array_container(operator.ne, x, y)
 
+    def logical_or(self, x, y):
+        return rec_multimap_array_container(cl_array.logical_or, x, y)
+
+    def logical_and(self, x, y):
+        return rec_multimap_array_container(cl_array.logical_and, x, y)
+
+    def logical_not(self, x):
+        return rec_map_array_container(cl_array.logical_not, x)
+
     # }}}
 
     # {{{ mathematical functions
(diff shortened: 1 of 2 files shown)


arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -259,6 +259,24 @@ def equal(self, x, y):
     def not_equal(self, x, y):
         return rec_multimap_array_container(operator.ne, x, y)
 
+    def logical_or(self, x, y):
+        def _rec(ary1, ary2):
+            if ary1.dtype != np.bool_ or ary2.dtype != np.bool_:
```
alexfikl, line 264: Doesn't pyopencl use `int8` for bools? Does this work as expected?

kaushikcfd: Thanks, yep. It was wrong earlier. Just fixed pyopencl to generate kernels to support all types. See https://github.com/inducer/pyopencl/pull/679.

arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -62,7 +63,7 @@ def _get_fake_numpy_linalg_namespace(self):
     def __getattr__(self, name):
         if name in self._pt_funcs:
             from functools import partial
-            return partial(rec_map_array_container, getattr(pt, name))
```
inducer, line 65: `multimap` is much more expensive than `map`. Why not separate unary and $n$-ary functions to avoid this expense when it is in fact avoiable?

kaushikcfd: Yep, fair. Will do.

kaushikcfd: See [d9d3fff](https://github.com/inducer/arraycontext/pull/219/commits/d9d3fff6608d2cbe4f8e6e1bed97b5677be1d565).

inducer:
Thanks!

PR #221, opened 2023-03-14
Add a pass to outline computations in a function

kaushikcfd:
Outlining is as simple as decorating a function with `@actx.outline`.

Draft because:
- [ ] Needs https://github.com/inducer/pytato/pulls.


majosm:
@inducer Do you know what's up with these doc errors? I don't get them when building locally.

Otherwise, this looks ready for review.

PR #228, opened 2023-04-26
single_version_only for actx.compile

matthiasdiener:


arraycontext/impl/pytato/__init__.py
```diff
@@ -612,9 +612,11 @@ def call_loopy(self, program, **kwargs):
 
         return call_loopy(program, processed_kwargs, entrypoint)
 
-    def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
+    def compile(self, f: Callable[..., Any],
+                single_version_only: bool = False) -> Callable[..., Any]:
         from .compile import LazilyPyOpenCLCompilingFunctionCaller
-        return LazilyPyOpenCLCompilingFunctionCaller(self, f)
+        return LazilyPyOpenCLCompilingFunctionCaller(self,
```
kaushikcfd, line 618: Maybe also handle the other sub-classes of BaseLazilyCompilingFunctionCaller?

arraycontext/impl/pytato/compile.py
```diff
@@ -262,6 +262,8 @@ class BaseLazilyCompilingFunctionCaller:
     program_cache: Dict["PMap[Tuple[Any, ...], AbstractInputDescriptor]",
                         "CompiledFunction"] = field(default_factory=lambda: {})
 
+    single_version_only: bool = False
```
kaushikcfd, line 265: I'm worried this might lead to developer errors. I would prefer if we don't default and ensure that the current version is passed during instantiation.

arraycontext/context.py
```diff
@@ -504,7 +504,8 @@ def clone(self: SelfType) -> SelfType:
             "setup-only" array context "leaks" into the application.
         """
 
-    def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
+    def compile(self, f: Callable[..., Any],
+                single_version_only: bool = False) -> Callable[..., Any]:
```
kaushikcfd, line 508: docs for single_version_only?

kaushikcfd (approved):
LGTM, thanks! We might need to move this interface to `ArrayContext.compile` or might lead to illegal codes for certain array contexts.

arraycontext/context.py
matthiasdiener: Added in 5ed25a59372c098976e55adb88ae302b3d75f82b

arraycontext/impl/pytato/__init__.py
matthiasdiener: Added in 5ed25a5

arraycontext/impl/pytato/compile.py
matthiasdiener: I removed the default in 5ed25a5

matthiasdiener: This fails in grudge now:

```
  File "wave-mpi-lazy.py", line 201, in main
    compiled_rhs = actx.compile(rhs)
                   ^^^^^^^^^^^^^^^^^
  File "/home/runner/work/arraycontext/arraycontext/mirgecom/src/grudge/grudge/array_context.py", line 416, in compile
    return _DistributedLazilyPyOpenCLCompilingFunctionCaller(self, f)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: BaseLazilyCompilingFunctionCaller.__init__() missing 1 required positional argument: 'single_version_only'
```

kaushikcfd: Hmm... Maybe then we introduce the default as you suggested, but have a deprecation period attached to it?

arraycontext/context.py
```diff
@@ -520,6 +521,8 @@ def compile(self, f: Callable[..., Any]) -> Callable[..., Any]:
         it may be called only once (or a few times).
 
         :arg f: the function executing the computation.
+        :arg single_version_only: If *True*, raise an error if *f* is compiled
+            more than once (due to different input argument types).
```
inducer, line 525: - This needs to explain quite a bit more context, such as by explaining what triggers a recompile.
- I'm not sure I like "version" as a noun here.

arraycontext/impl/pytato/compile.py
```diff
@@ -324,7 +325,10 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
         try:
             compiled_f = self.program_cache[arg_id_to_descr]
```
inducer, line 326: Can we save ourselves having to find `arg_id_to_descr`?

inducer:
@matthiasdiener Could you measure the impact this has on launch overhead, for mirgecom-like kernels? This would help in deciding how to prioritize this.

arraycontext/impl/pytato/compile.py
matthiasdiener: If I remember correctly, this was the main goal of implementing single_version_only?

What do you think of the approach in [d8071ea](https://github.com/inducer/arraycontext/pull/228/commits/d8071eaabbaf9684773caa2976b99b6dcc8c93ef)?

matthiasdiener:
> @matthiasdiener Could you measure the impact this has on launch overhead, for mirgecom-like kernels? This would help in deciding how to prioritize this.

With the current version (as of b3c69c7), this has a measurable impact on the time step time for smaller examples:

Sod, 1 rank
| Config | Time per time step [s]| 
| -------- | -------- |
| `single_version_only=False` (default) | 0.0265 |
| `single_version_only=True` | 0.0249 |

Unfortunately, I did not see any meaningful difference in time step time for large drivers, such as `ks3d`.

PR #229, opened 2023-05-05
Copy `axes` and `tags` in `zeros_like`/`full_like`

majosm:
Am I doing this right? (As far as I could tell, `pt.zeros`/`pt.full` don't accept `axes` or `tags` directly.)

This makes all of the RHS "uninferred axes" warnings go away in my Y3 prediction `smoke_test` run.

cc @anderson2981

### DIFF
--- a/arraycontext/impl/pytato/fake_numpy.py
+++ b/arraycontext/impl/pytato/fake_numpy.py
@@ -73,7 +73,8 @@ class PytatoFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
 
     def zeros_like(self, ary):
         def _zeros_like(array):
-            return self._array_context.zeros(array.shape, array.dtype)
+            return self._array_context.zeros(
+                array.shape, array.dtype).copy(axes=array.axes, tags=array.tags)
 
         return self._array_context._rec_map_container(
             _zeros_like, ary, default_scalar=0)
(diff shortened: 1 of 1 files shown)


inducer:
LGTM!

inducer:
(And thanks for getting that done.)

majosm:
Looks like this change is making the stock prediction driver (not my slightly-modified version) fail CI: https://github.com/illinois-ceesd/mirgecom/actions/runs/4896641880/jobs/8744004554. @MTCam confirmed by swapping arraycontext versions. Any idea what could be causing this @inducer, @kaushikcfd?

inducer:
I think I understand how this could happen, and I think one way to obviate the issue for now would be to delete the [offending assert](https://github.com/majosm/meshmode/blob/fbd81e419d7d12f41e4af9de93236f26636ccf3c/meshmode/array_context.py#L1249).

`FEMEinsumTag`s are supposed to come only into existence as part of the transform path, and so it's a problem if they exist "before they're even supposed to be created". The bit of code that the `assert` is part of ensures that only "input-type" arrays (`Placeholder`, `DataWrapper`) have this issue. This change to `zeros_like` permits that tag to "jump" to `IndexLambda`s, too. I suppose that the cleaner solution would ultimately be to ensure that the tag does not escape the transform process, but deleting the `assert` should do the trick for now.

@kaushikcfd, LMK if I've got this wrong.

kaushikcfd:
That assessment seems correct. The FEMEinsumTag was a broken abstraction, the newest version of `kaushikcfd@meshmode` is based on the implementation in #217 which gets rid of them.

PR #234, opened 2023-06-19
tests: fix comparison on numpy 1.25

alexfikl:
Before comparing two `ndarray([DOFArray, DOFArray])` would just return `ndarray([False, False])`, but it doesn't work anymore. Couldn't find an explanation in the release notes as to why that changed.

I changed it to use `actx.np.equal`, but I'm not sure if this was supposed to test `__eq__`?

### DIFF
--- a/test/test_arraycontext.py
+++ b/test/test_arraycontext.py
@@ -1311,7 +1311,8 @@ def test_container_equality(actx_factory):
     dc2 = MyContainer(name="yoink", mass=ary_dof, momentum=None, enthalpy=None)
     assert dc != dc2
 
-    assert isinstance(bcast_dc_of_dofs == bcast_dc_of_dofs_2, MyContainerDOFBcast)
+    assert isinstance(actx.np.equal(bcast_dc_of_dofs, bcast_dc_of_dofs_2),
+                      MyContainerDOFBcast)
 
 # }}}
 


inducer:
Thanks for working on this. The following:
```python
import numpy as np


class MyContainer:
    def __eq__(self, other):
        return self

    def __bool__(self):
        raise ValueError("not defined")


a = np.zeros(2, dtype=object)
a[0] = MyContainer()
a[1] = MyContainer()

b = np.zeros(2, dtype=object)
b[0] = MyContainer()
b[1] = MyContainer()

a == b
```
says
```
/home/andreas/tmp/numpy-obj-array-equal.py:20: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
  a == b
```
so it seems that this may have been an intentional change.

inducer:
I think this may be the bit about `==` in https://numpy.org/doc/stable/release/1.25.0-notes.html#expired-deprecations.

inducer:
This means that `==` is no longer really a thing for array containers involving numpy object arrays, which I guess is just as well. I think your change is more or less the only option here.

alexfikl:
> I think this may be the bit about `==` in [numpy.org/doc/stable/release/1.25.0-notes.html#expired-deprecations](https://numpy.org/doc/stable/release/1.25.0-notes.html#expired-deprecations).

Oh yeah, I totally missed that. Thanks for hunting it down!

PR #235, opened 2023-06-20
add NumpyArrayContext

matthiasdiener:
Continuation of #93, #190.

_Please squash_


matthiasdiener:
I think this is ready for review. With a few small changes to the mirgecom example drivers, I could get them to run with the NumpyArrayContext in this PR.

PR #238, opened 2023-07-24
_dag_to_transformed_pytato_prg: use _DEFAULT_LOOPY_OPTIONS

matthiasdiener:
the options are currently identical, but this makes future changes easier

PR #239, opened 2023-07-24
disable array bounds check when optimized

matthiasdiener:
Needs:
- [x] #238.

### DIFF
--- a/arraycontext/loopy.py
+++ b/arraycontext/loopy.py
@@ -27,6 +27,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+import sys
+
 import numpy as np
 
 import loopy as lp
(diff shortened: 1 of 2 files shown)


arraycontext/loopy.py
```diff
@@ -41,7 +43,9 @@
 
 _DEFAULT_LOOPY_OPTIONS = lp.Options(
         no_numpy=True,
-        return_dict=True)
+        return_dict=True,
+        enforce_array_accesses_within_bounds="no_check" if sys.flags.optimize
```
inducer, line 47: I'm a bit hesitant about this, because I feel it encourages everyone to run with `-O` all the time, negating the benefit of possible runtime checks.

 I think a possibly better approach might be to just turn off bounds checking for kernels generated by pytato wholesale, as long as no hand-written loopy kernels appear in the DAG.

matthiasdiener:
Closing in favor of https://github.com/inducer/pytato/pull/449.

PR #240, opened 2023-07-26
Port to new TranslationUnit.executor

alexfikl:
This caches the executor in `_loopy_transform_cache`. Is that ok?


inducer:
Thanks!

I was working on also using the executor stuff in the lazy codepath, which required [a bit of a detour](https://github.com/inducer/pytato/pull/450). Otherwise, I already have analogous changes lined up (with the addition of also using the executor business on `freeze`).

Caching in `_loopy_transform_cache` is just fine: That's specific to an array context, which is itself specific to a CL context, so there's no risk of holding an uninvited reference to a CL context.

Still, I'll close this in favor of my version of this. Thanks again for making this, and sorry about the duplicate work.

inducer:
Superseded by #241.

PR #251, opened 2024-02-16
Implement CupyArrayContext

matthiasdiener:


test/test_arraycontext.py
```diff
@@ -1245,6 +1251,9 @@ def scale_and_orthogonalize(alpha, vel):
 
 def test_actx_compile(actx_factory):
     actx = actx_factory()
+    if isinstance(actx, CupyArrayContext):
+        pytest.skip("CupyArrayContext does not support object arrays")
+
```
matthiasdiener, line 1256: @inducer: We had briefly discussed adding support for object arrays via a wrapper (?), but I'm not sure how that should look like.

arraycontext/pytest.py
```diff
@@ -224,6 +224,26 @@ def __str__(self):
...
+
+    def __call__(self):
+        from arraycontext import CupyArrayContext
+        return CupyArrayContext()
+
+    def __str__(self):
+        import cupy  # pylint: disable=import-error
+        d = cupy.cuda.runtime.getDeviceProperties(cupy.cuda.Device())
+        name = d["name"].decode("utf-8")
+        return f"<CupyArrayContext> on {cupy.cuda.Device()}:{name}"
```
matthiasdiener, line 244: This is the only part of the code that give me pause for supporting something like [cupy-rocm](https://docs.cupy.dev/en/latest/install.html#using-cupy-on-amd-gpu-experimental).

matthiasdiener:
This is ready for a first look @inducer

arraycontext/pytest.py
matthiasdiener: cupy-rocm seems to be completely out-of-date, so probably no big deal.

test/test_arraycontext.py
matthiasdiener: Turns out this was caused by the incomplete array container support, fixed in [e9c42dd](https://github.com/inducer/arraycontext/pull/251/commits/e9c42dd193c2c29f1c9322f7cac49ae81af8302e). Thanks to @majosm for the help!

matthiasdiener:
This is ready for review @inducer

matthiasdiener:
With the recent changes, the CupyArrayContext works fine for a ks3d simulation (with https://github.com/illinois-ceesd/drivers_y3-prediction/pull/73).

Performance (1 rank ks3d on porter):

- `NumpyArrayContext`: 28.4s per step
- `CupyArrayContext`: 4.0s per step

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+class CupyArrayContext(ArrayContext):
+    """
+    An :class:`ArrayContext` that uses :class:`cupy.ndarray` to represent arrays.
+
+    .. automethod:: __init__
+    """
+
+    _loopy_transform_cache: dict[lp.TranslationUnit, lp.ExecutorBase]
+
+    def __init__(self) -> None:
```
inducer, line 74: Would it be useful to specify a device on array context creation.

https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.Device.html#cupy-cuda-device

Is device usage "sticky" per created array? If not, what happens if a device is currently default that doesn't have the data?

.github/workflows/ci.yml
```diff
@@ -52,6 +56,9 @@ jobs:
                 curl -L -O https://tiker.net/ci-support-v0
                 . ./ci-support-v0
 
+                CONDA_ENVIRONMENT=.test-conda-env-py3.yml
+                echo "- cupy" >> "$CONDA_ENVIRONMENT"
```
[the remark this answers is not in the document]
matthiasdiener: If cupy is not installed, these tests will fail, e.g. mypy with `arraycontext/impl/cupy/__init__.py:57: error: Cannot find implementation or library stub for module named "cupy"  [import-not-found]`. We could `type: ignore` these, but at least in the case of pylint we would have to annotate every `import cupy` line with `# pylint: disable=import-error`. Not sure which way is better.

arraycontext/impl/cupy/__init__.py
matthiasdiener: I've added an optional argument to `__init__` to specify the device in ab8266d. Is that what you had in mind?


> Is device usage "sticky" per created array? If not, what happens if a device is currently default that doesn't have the data?

I don't think it is sticky, see https://docs.cupy.dev/en/stable/user_guide/basic.html#current-device for some context:

> Note
> If the array’s device and the current device mismatch, CuPy functions try to establish [peer-to-peer memory access](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#peer-to-peer-memory-access) (P2P) between them so that the current device can directly read the array from another device. Note that P2P is available only when the topology permits it. If P2P is unavailable, such an attempt will fail with ValueError.

.github/workflows/ci.yml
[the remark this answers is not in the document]
matthiasdiener: Done in [79b0bc3](https://github.com/inducer/arraycontext/pull/251/commits/79b0bc37e75d4f58fb9c6f98323fb9a3b158476a) et al., and seems to work: https://gitlab.tiker.net/inducer/arraycontext/-/jobs/804938

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+        _, result = executor(**kwargs)
+
+        return result
+
+    def freeze(self, array):
+        import cupy as cp
+
+        def _freeze(ary):
+            return cp.asnumpy(ary)
```
alexfikl, line 150: Does `cupy` have something like `jax.block_until_ready` or `pyopencl.finish`? This transfers the memory back to the host, right?

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        import cupy as cp
+
+        def _thaw(ary):
+            return cp.array(ary)
+
+        return with_array_context(rec_map_array_container(_thaw, array), actx=self)
+
+    # }}}
+
+    def transform_loopy_program(self, t_unit):
```
alexfikl, line 164: This should be deleted.

alexfikl (commented):
Was curious a bit about this and left some comments. Hope that's ok!

What's the use of adding a `cupy` backend? I thought eager ones don't really give good performance and we already have both the `pyopencl` and the `jax` ones that can use CUDA (behind some abstraction layers, at least).

.github/workflows/ci.yml
```diff
@@ -40,6 +40,10 @@ jobs:
             run: |
                 USE_CONDA_BUILD=1
                 curl -L -O https://gitlab.tiker.net/inducer/ci-support/raw/master/prepare-and-run-pylint.sh
+
+                CONDA_ENVIRONMENT=.test-conda-env-py3.yml
+                echo "- cupy" >> "$CONDA_ENVIRONMENT"
```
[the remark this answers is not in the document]
matthiasdiener: Sure, but I'll wait for @inducer's decision regarding whether this dependency should be added for mypy at all.

matthiasdiener:
> Was curious a bit about this and left some comments. Hope that's ok!
> 
> What's the use of adding a `cupy` backend? I thought eager ones don't really give good performance and we already have both the `pyopencl` and the `jax` ones that can use CUDA (behind some abstraction layers, at least).

Thank you for the review! 

I don't have experience with jax, and we generally don't run mirgecom with it. Regarding the pyopencl-eager backend(s), it does not support all operations that we need (e.g., broadcasting), so we can't run all our drivers with it. This only leaves us with numpy or pytato backends. pytato has generally high compile times, so it's sometimes not great for debugging etc. Numpy time step times are often much slower than with cupy, as indicated above.

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+    array_types = (CupyNonObjectArray,)
+
+    def _get_fake_numpy_namespace(self):
+        from .fake_numpy import CupyFakeNumpyNamespace
+        return CupyFakeNumpyNamespace(self)
+
+    # {{{ ArrayContext interface
+
+    def clone(self):
+        return type(self)()
```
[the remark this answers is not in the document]
matthiasdiener: Good catch, thanks!

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def __init__(self, device: int | None = None) -> None:
+        super().__init__()
+        self._loopy_transform_cache = {}
+
+        if device is not None:
+            import cupy as cp
+            cp.cuda.runtime.setDevice(device)
+
+    array_types = (CupyNonObjectArray,)
```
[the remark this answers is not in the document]
matthiasdiener: Not opposed, but my goal was to keep this implementation as close to the numpy actx as possible.

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        from .fake_numpy import CupyFakeNumpyNamespace
+        return CupyFakeNumpyNamespace(self)
+
+    # {{{ ArrayContext interface
+
+    def clone(self):
+        return type(self)()
+
+    @overload
+    def from_numpy(self, array: np.ndarray) -> Array:
```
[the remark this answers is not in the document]
matthiasdiener: Same as above - not opposed, but my goal was to keep this implementation as close to the numpy actx as possible.

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+            return cp.asnumpy(ary)
+
+        return with_array_context(rec_map_array_container(_to_numpy, array),
+                                  actx=None)
+
+    def call_loopy(
+                self,
+                t_unit: lp.TranslationUnit, **kwargs: Any
+            ) -> dict[str, Array]:
+        t_unit = t_unit.copy(target=lp.ExecutableCTarget())
```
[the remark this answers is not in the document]
matthiasdiener: You're right, loopy support is probably futile. Removed in ce33f23

arraycontext/impl/cupy/__init__.py
matthiasdiener: it does - we could synchronize on the (default) stream, for example. Not sure if it's worth the extra effort though.

arraycontext/impl/cupy/__init__.py
matthiasdiener: Changed it to raise a NotImplementedError in ce33f23.

arraycontext/impl/cupy/__init__.py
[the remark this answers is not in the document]
matthiasdiener: It is in the numpy actx as well as far as I can see. mypy does complain locally without it: 

```
arraycontext/impl/cupy/__init__.py:94: error: Signature of "from_numpy" incompatible with supertype "ArrayContext"  [override]
arraycontext/impl/cupy/__init__.py:94: note:      Superclass:
arraycontext/impl/cupy/__init__.py:94: note:          @overload
arraycontext/impl/cupy/__init__.py:94: note:          def from_numpy(self, array: ndarray[Any, Any]) -> Array
arraycontext/impl/cupy/__init__.py:94: note:          @overload
arraycontext/impl/cupy/__init__.py:94: note:          def [ContainerOrScalarT: ArrayContainer | int | float | complex | generic] from_numpy(self, array: ContainerOrScalarT) -> ContainerOrScalarT
arraycontext/impl/cupy/__init__.py:94: note:      Subclass:
arraycontext/impl/cupy/__init__.py:94: note:          def from_numpy(self, array: ndarray[Any, Any] | ArrayContainer | int | float | complex | generic) -> Array | ArrayContainer | int | float | complex | generic
```

arraycontext/pytest.py
```diff
@@ -227,6 +227,26 @@ def __str__(self):
         return "<PytatoJAXArrayContext>"
 
 
+class _PytestCupyArrayContextFactory(PytestArrayContextFactory):
+    @classmethod
+    def is_available(cls) -> bool:
+        try:
+            import cupy  # type: ignore[import-untyped]  # noqa: F401
```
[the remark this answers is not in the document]
alexfikl: Maybe add that to `pyproject.toml` then? `follow_untyped_imports = false` (not sure that's the option).

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+    _loopy_transform_cache: dict[lp.TranslationUnit, lp.ExecutorBase]
+
+    def __init__(self, device: int | None = None) -> None:
+        super().__init__()
+        self._loopy_transform_cache = {}
+
+        if device is not None:
+            import cupy as cp
+            cp.cuda.runtime.setDevice(device)
```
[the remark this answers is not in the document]
alexfikl: > I think having the possibility of choosing a device makes sense in general, particularly in multi-GPU systems. I don't know how jax handles these systems, but pyopencl actx does support it through the CommandQueue argument in their constructors.

My understanding is that `jax` handles it the same: the user gets to set the backend globally either through some call or some environment variable. `pyopencl` is different because multiple queues can exist at the same time.

I'm not against passing the device in here, since it doesn't seem to hurt anything, but was just wondering if there's any benefit (i.e. the calling code can just add a line to set it).

arraycontext/impl/cupy/__init__.py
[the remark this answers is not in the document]
alexfikl: Fair enough, was just weird to see a class variable after `__init__` :))

arraycontext/impl/cupy/__init__.py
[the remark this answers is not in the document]
alexfikl: Well that's quite annoying. Does it not inherit the overloads from the base class at all? i.e. every implementation will have to copy past these empty overloads? :(

arraycontext/impl/cupy/__init__.py
alexfikl: I was mostly worried that copying the memory from the device and back on freeze+thaw could hurt performance? Is that not an issue?

Might be worth adding a comment here to explain why this choice was made.

alexfikl:
> I don't have experience with jax, and we generally don't run mirgecom with it. Regarding the pyopencl-eager backend(s), it does not support all operations that we need (e.g., broadcasting), so we can't run all our drivers with it. This only leaves us with numpy or pytato backends. pytato has generally high compile times, so it's sometimes not great for debugging etc. Numpy time step times are often much slower than with cupy, as indicated above.

Ah.. broadcasting.. the bane of everyone's existence! Fair enough :(

arraycontext/impl/cupy/__init__.py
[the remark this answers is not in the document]
alexfikl: Although, if the device setting is a global option, I'm not sure this makes sense? e.g. something like
```
# make some arraycontext on device 1
# do something with actx on device 1
# make an array context on device 2
# clone array context with device 1 --> device 2 is suddenly out
```
I imagine nobody will muck around with devices like that though :))

arraycontext/impl/cupy/__init__.py
inducer: Wait, what? The text you quoted seems to imply that cupy *will* attempt automatic migration.

My read on this situation is that it's the "current device" setting in cupy is the only thing that matters, there's no point in exposing this as a constructor parameter since it won't "stick" with the created arrays.

matthiasdiener: > Wait, what? The text you quoted seems to imply that cupy _will_ attempt automatic migration.

Hmm, where do you see the implication of migration? My understanding is that it will access the memory on the other device, but not _migrate_ the memory (a bit like a remote NUMA access).

**On Porter:**
```python
>>> import cupy as cp
>>> a = cp.array([1,2,3])
>>> a.device
<CUDA Device 0>
>>> cp.cuda.runtime.setDevice(1)
>>> a.device
<CUDA Device 0>
>>> b = a + 4
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "cupy/_core/core.pyx", line 1275, in cupy._core.core._ndarray_base.__add__
  File "cupy/_core/core.pyx", line 1699, in cupy._core.core._ndarray_base.__array_ufunc__
  File "cupy/_core/_kernel.pyx", line 1286, in cupy._core._kernel.ufunc.__call__
  File "cupy/_core/_kernel.pyx", line 159, in cupy._core._kernel._preprocess_args
  File "cupy/_core/_kernel.pyx", line 130, in cupy._core._kernel._preprocess_arg
  File "cupy/_core/_kernel.pyx", line 120, in cupy._core._kernel._check_peer_access
ValueError: The device where the array resides (0) is different from the current device (1). Peer access is unavailable between these devices.
```

**On Lassen:**

```python
>>> import cupy as cp
>>> a = cp.array([1,2,3])
>>> a.device
<CUDA Device 0>
>>> cp.cuda.runtime.setDevice(1)
>>> a.device
<CUDA Device 0>
>>> b = a + 4
<stdin>:1: PerformanceWarning: The device where the array resides (0) is different from the current device (1). Peer access has been activated automatically.
>>> b
array([5, 6, 7])
>>> b.device
<CUDA Device 1>
>>> a.device
<CUDA Device 0>
```

inducer: OIC, nvm, I misunderstood this, too. Still, I think wrapping "set current device" via the constructor isn't helpful, since it won't exercise positive control over array placement. That is, unless we intend to also set the active device before every computation. (Which, in actuality, we kind of can't, at least not without fully wrapping all cupy functionality.)

matthiasdiener: Ok, I'll revert the relevant commits.

arraycontext/pytest.py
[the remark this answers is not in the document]
matthiasdiener: Not sure if that's better - in the current way, I think it is clearer which package actually has missing types.

arraycontext/impl/cupy/__init__.py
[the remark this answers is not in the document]
matthiasdiener: I reverted the relevant commits.

arraycontext/pytest.py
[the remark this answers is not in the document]
alexfikl: I meant to add them under a `tool.mypy.overrides` section to avoid littering the code, but I'm fine with keeping a few of them if you think it's clearer.

arraycontext/impl/cupy/__init__.py
matthiasdiener: Removed in 4d429f8

arraycontext/impl/cupy/__init__.py
matthiasdiener: I did try running mirgecom with a nonblocking `cp.asnumpy` plus a stream synchronization, the time step time was not affected at all. I added a brief comment in 69e2133

README.rst
```diff
@@ -16,6 +16,7 @@ code to work with all of them? No problem! Comes with pre-made array context
 implementations for:
 
 - numpy
+- cupy
```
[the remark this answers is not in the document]
matthiasdiener: Added in [f83ee9c](https://github.com/inducer/arraycontext/pull/251/commits/f83ee9c2a15b29fcfa16e64400ca18b7ea2a7390)

matthiasdiener:
This is ready for another review @inducer @alexfikl

README.rst
```diff
@@ -25,7 +25,7 @@ implementations for:
 - Profiling
```
matthiasdiener, line 25: Out of curiosity, what are the implementations for debugging/profiling?

arraycontext/impl/cupy/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class CupyArrayContext(ArrayContext):
+    """
+    An :class:`ArrayContext` that uses :class:`cupy.ndarray` to represent arrays.
+
+    .. automethod:: __init__
+    """
+
+    _loopy_transform_cache: dict[lp.TranslationUnit, lp.ExecutorBase]
```
alexfikl, line 71: This can also go :grin:

alexfikl (approved):
Left a little nitpick, but otherwise this looks good to me! Thanks!

arraycontext/impl/cupy/__init__.py
matthiasdiener: Good catch, thanks! d9ce8d5

README.rst
alexfikl: I don't think there are any at the moment (?). Unless passing a `CommandQueue` with profiling enabled counts there..

It would be nice to have some fancier profiling for things though :grin:

PR #252, opened 2024-03-01
jax: update config import

alexfikl:
The latest version of jax updated the location of the config object:
https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-25-feb-26-2024

I think this should work in older versions too, since `jax.config.update` has been there (just the explicit `jax.config.config` object moved)


inducer:
Thx!

Issue #247, opened 2023-09-26
Loopy a hard dependency?


kaushikcfd:
Is there a specific reason why Loopy should be a hard dependency[^1]? My understanding was `transform_loopy_program` will get kicked in only when the implementation demands.

Thanks!

[^1]: https://github.com/inducer/arraycontext/blob/main/setup.py#L46

inducer:
I would be sympathetic to removing it. The one conceivable reason is that `call_loopy` is nominally part of the actx interface, but I think that should be considered an optional part, and it could conceivably be made a function, rather than a method, possibly via `singledispatch`. As you know, there's precedent (Jax) that doesn't support it.

PR #267, opened 2024-07-09
Move project metadata and remove setup.py

alexfikl:


pyproject.toml
```diff
@@ -1,5 +1,77 @@
...
+    "jax>=0.4",
+]
+pyopencl = [
+    "islpy>=2024.1",
+    "loopy>=2024.1",
+    "pyopencl>=2024.1",
+]
+pytato = [
+    "pytato>=2021.1",
+]
```
alexfikl, line 49: I added all of the various array context implementations as optional dependencies here. Let me know if you prefer them somewhere else!

alexfikl: Potentially "fixes" #247 if left like this :grin:

inducer: Putting those there seems OK. As for #247, there's still `call_loopy`, which is a bit questionable.

alexfikl: Yeah, the code itself is still quite dependent on loopy, at least morally :(

PR #268, opened 2024-07-13
container: fix warning for cls_has_array_context_attr

alexfikl:
It had a `{cls}.array_context` in there that became `<class 'meshmode.dof_array.DOFArray'>.array_context`. Changed now to just use the nicer `DOFArray.array_context`, since the full path of the class is mentioned above.


inducer:
Thx!

PR #275, opened 2024-08-05
Ignore mypy false positive

alexfikl:
Haven't managed to track down the actual issue, but the error here goes away if `tag_dataclass` is replaced with just `dataclass`. mypy has multiple `dataclass_transform` issues still open, so it's likely not quite ironed out..

Gitlab is passing though: https://gitlab.tiker.net/fikl2/arraycontext/-/pipelines/575042

### DIFF
--- a/arraycontext/impl/pytato/__init__.py
+++ b/arraycontext/impl/pytato/__init__.py
@@ -101,7 +101,9 @@ def _preprocess_array_tags(tags: ToTagSetConvertible) -> FrozenSet[Tag]:
                     "was already present.", stacklevel=1)
 
         tags = (
-                (tags | frozenset({PrefixNamed(name_hint.name)}))
+                # NOTE: mypy seems to be confused with `dataclass_transform` so
+                # these don't always work as expected for now..
+                (tags | frozenset({PrefixNamed(name_hint.name)}))  # type: ignore[call-arg]
                 - {name_hint})
 
     return tags


alexfikl:
@inducer Dug into this a bit more and it seems like the problem is that we pin `python_version = "3.8"` but the `dataclass_transform` is a python 3.11 feature, so mypy doesn't know what to do about it. Making the same pin in `pytato` also yells about all the classes using `tag_dataclass`.

From what I can tell the possible fixes are: bump the mypy python version or ignore it harder (i.e. add a temporary `unused-ignore` too). What would you prefer?

inducer:
Thanks for digging into this! So is the overall issue intermittent? (Given that it passed here when it, IIUC, wasn't expected to?) Is there a mypy issue we can point to? To my mind, a Py3.8 codebase that imports 3.11 stuff from `typing_extensions` should be perfectly fine?

alexfikl:
> So is the overall issue intermittent? (Given that it passed here when it, IIUC, wasn't expected to?) 

It's not intermittent. It's seems to pass consistently on Github in the conda env and fail consistently on Gitlab in the pip virtual env. I see the same failures locally as on Gitlab.

> To my mind, a Py3.8 codebase that imports 3.11 stuff from `typing_extensions` should be perfectly fine?

Yeah, agreed, not sure why mypy is confused about it here..

alexfikl:
Bleah, I think I know what it doesn't like. Gitlab seems to install pytools from git and Github from conda (or something?).

This simple reproducer fails for me. If you only import `typing_extensions`, it works just fine :( So I guess mypy is confused by the `try.. except`?

```python
from __future__ import annotations

from dataclasses import dataclass
from typing import TypeVar

try:
    from typing import dataclass_transform
except ImportError:
    from typing_extensions import dataclass_transform

T = TypeVar("T")


@dataclass_transform(eq_default=True, frozen_default=True)
def tag(cls: type[T]) -> type[T]:
    return dataclass(init=True, frozen=True, eq=True, repr=True)(cls)


@tag
class Tag:
    pass


@tag
class UniqueTag(Tag):
    pass


@tag
class _BaseNamedTag(Tag):
    pass


@tag
class SomeTag(_BaseNamedTag):
    name: str


t1 = Tag()
t2 = SomeTag("name")
```

alexfikl:
I'll go ahead and close this since it just needs a new pytools with the fix.

PR #278, opened 2024-08-13
Forbid setting `force_device_scalars=False`

majosm:
Deprecated since 2022, AFAIK no longer needed?

To-do:
- [x] Merge https://github.com/inducer/meshmode/pull/429
- [x] Merge https://github.com/inducer/grudge/pull/358

### DIFF
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -84,7 +84,7 @@ class PyOpenCLArrayContext(ArrayContext):
             queue: pyopencl.CommandQueue,
             allocator: Optional[pyopencl.tools.AllocatorBase] = None,
             wait_event_queue_length: Optional[int] = None,
-            force_device_scalars: bool = False) -> None:
+            force_device_scalars: Optional[bool] = None) -> None:
         r"""
         :arg wait_event_queue_length: The length of a queue of
             :class:`~pyopencl.Event` objects that are maintained by the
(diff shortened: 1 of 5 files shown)


majosm:
Looks like the default value of `force_device_scalars` will need to be changed in the meshmode/grudge array context constructors before this can be merged. I'll wait until this gets a tentative go-ahead before doing that.

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -303,8 +296,7 @@ def call_loopy(self, t_unit, **kwargs):
 
     def clone(self):
         return type(self)(self.queue, self.allocator,
-                wait_event_queue_length=self._wait_event_queue_length,
-                force_device_scalars=self._force_device_scalars)
+                wait_event_queue_length=self._wait_event_queue_length)
```
alexfikl, line 299: I think this needs a `force_device_scalars=None` to stay, so that subclasses that still have that argument get the expected result on clone? (hopefully also means no immediate changes in grudge / meshmode)

alexfikl (commented):
This has been deprecated for a long time now, so I'm definitely in favor! :tada:

arraycontext/impl/pyopencl/__init__.py
alexfikl: Might also need to keep `self._force_device_scalars = None` in `__init__` (or as a deprecated property).

inducer:
Very much in favor as well!

arraycontext/impl/pyopencl/__init__.py
majosm: Thanks! Yeah, looks like `self._force_device_scalars` needs to stick around for now, since subclasses can/do use it. I attempted to address the remaining `clone()` compatibility in https://github.com/inducer/arraycontext/pull/278/commits/88d33413c8b3885aaeb9f6cb0e972d53667a1c29.

majosm:
Ruff errors appear to be unrelated, so I think this is ready for a look. https://github.com/inducer/meshmode/pull/429 and https://github.com/inducer/grudge/pull/358 should probably get merged before this.

arraycontext/impl/pyopencl/__init__.py
alexfikl: That looks good to me! Thanks!

It seems like all the downstream tests are passing now too :grin:

PR #280, opened 2024-08-25
Introduce `Bcast` object-ified broacasting rules

inducer:
See discussion in #190 for context, where this was fairly controversial.


inducer:
See #310 for attempt 2.

PR #281, opened 2024-08-30
add minimum, fmin

matthiasdiener:
_Please squash_


inducer:
Thanks! The lint fixes (slightly different from yours) will come via #190, hopefully soon. This will go in after that.

PR #282, opened 2024-09-25
add device<->host transfer mappers

matthiasdiener:
Followup of https://github.com/inducer/pytato/pull/548

TODOs:
- [ ] add to docs?

### DIFF
--- a/arraycontext/impl/pytato/utils.py
+++ b/arraycontext/impl/pytato/utils.py
@@ -25,6 +25,8 @@ THE SOFTWARE.
 
 from typing import TYPE_CHECKING, Any, Dict, Mapping, Set, Tuple
 
+from pyopencl import CommandQueue
+from pyopencl.tools import AllocatorBase
 from pytato.array import (
     AbstractResultWithNamedArrays,
     Array,
@@ -36,7 +38,7 @@ from pytato.array import (
     make_placeholder,
 )
 from pytato.target.loopy import LoopyPyOpenCLTarget
-from pytato.transform import CopyMapper
+from pytato.transform import ArrayOrNames, CopyMapper
 from pytools import UniqueNameGenerator, memoize_method
 
 from arraycontext.impl.pyopencl.taggable_cl_array import Axis as ClAxis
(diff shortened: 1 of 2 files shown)


arraycontext/impl/pytato/utils.py
```diff
@@ -124,4 +126,60 @@ def get_loopy_target(self) -> "lp.PyOpenCLTarget":
...
+    def __init__(self, queue: CommandQueue, allocator: AllocatorBase = None) -> None:
+        super().__init__()
+        self.queue = queue
+        self.allocator = allocator
+
+    def map_data_wrapper(self, expr: DataWrapper) -> Array:
+        import arraycontext.impl.pyopencl.taggable_cl_array as tga
+        if isinstance(expr.data, tga.TaggableCLArray):
+            data = expr.data.get()
+            return DataWrapper(
```
matthiasdiener, line 164: @majosm asked in https://github.com/inducer/pytato/pull/548#discussion_r1773622516

> Can we make this return an instance of a new DataWrapper type that can __hash__ by data instead of id? That would remove the need for data wrapper deduplication, if I understand the intended use of these mappers correctly.

matthiasdiener: Not sure; in that case, we probably should probably also override the `EqualityComparer`, otherwise we would get a lot of hash collisions when caching.

matthiasdiener:
This is ready for a first review; I'm not sure what to do about the pylint errors.

arraycontext/impl/pytato/utils.py
```diff
@@ -124,4 +125,61 @@ def get_loopy_target(self) -> "lp.PyOpenCLTarget":
...
+                             "is already on the device")
+
+        new_dw = self.actx.from_numpy(expr.data)
+        assert isinstance(new_dw, DataWrapper)
+        return DataWrapper(
+            data=new_dw.data,
+            shape=expr.shape,
+            axes=expr.axes,
+            tags=expr.tags,
+            non_equality_tags=expr.non_equality_tags)
```
matthiasdiener, line 150: Ideally, this should just do `return self.actx.from_numpy(expr.data).tagged(expr.tags)` I think, but there seems to be no way to transfer the `non_equality_tags` in that case.

matthiasdiener: Added as a comment

inducer:
Please do add these to the docs, and then this should be ready to go.

matthiasdiener:
> Please do add these to the docs, and then this should be ready to go.

Added in [b20d359](https://github.com/inducer/arraycontext/pull/282/commits/b20d3592efef0b8f28388585c2a9971e0572ad39)

PR #284, opened 2024-10-14
Fix new mypy 1.12 errors

alexfikl:


arraycontext/container/dataclass.py
```diff
@@ -83,14 +83,15 @@ def is_array_field(f: Field) -> bool:
                         f"Field '{f.name}' union contains non-array container "
                         "arguments. All arguments must be array containers.")
 
+        if isinstance(f.type, str):
```
alexfikl, line 86: `f.type` is not correctly a `Union[type[Any], str]`, which disagreed with `is_array_type` below.

inducer: Sounds good. In the short-ish term, we'll probably need to handle this case better, too. Maybe just tossing it to `parse` might be enough? (though some of the identifiers may not resolve?) At any rate, a discussion for a different day.

PR #285, opened 2024-11-05
Bump Python to 3.10

alexfikl:


arraycontext/impl/numpy/fake_numpy.py
```diff
@@ -143,13 +145,12 @@ def array_equal(self, a: ArrayOrContainer, b: ArrayOrContainer) -> Array:
...
-                    np.logical_and,
+            return np.logical_and.reduce(
                     [(true_ary if kx_i == ky_i else false_ary)
-                        and self.array_equal(x_i, y_i)
+                        and cast(np.ndarray, self.array_equal(x_i, y_i))
                         for (kx_i, x_i), (ky_i, y_i)
-                        in zip(serialized_x, serialized_y)],
-                    true_ary)
+                        in zip(serialized_x, serialized_y, strict=True)],
+                    initial=true_ary)
```
alexfikl, line 153: Had to change this a bit because `mypy` was unhappy.
```
Argument 1 to "reduce" has incompatible type "_UFunc_Nin2_Nout1[Literal['logical_and'], Literal[20], Literal[True]]"; expected "Callable[[ndarray[Any, dtype[Any]], ndarray[Any, dtype[Any]] | Array], ndarray[Any, dtype[Any]]]"  [arg-type]
arraycontext/impl/numpy/fake_numpy.py:149: note: "_UFunc_Nin2_Nout1[Literal['logical_and'], Literal[20], Literal[True]].__call__" has type overloaded function
```
Not quite sure why this doesn't show up on main too..

PR #288, opened 2024-11-15
Improve dataclass container

alexfikl:
This updates `dataclass_container` with some new typing things:
* Apparently `A | B` shows up as `types.UnionType`, not `Union`.
* Also `tuple[A, B]` shows up as `types.GenericAlias`, which is not a subclass of `typing._BaseGenericAlias`.


inducer:
Thanks!

PR #289, opened 2024-11-24
Fix ruff re errors

alexfikl:


arraycontext/impl/pytato/utils.py
```diff
@@ -190,7 +190,7 @@ def map_data_wrapper(self, expr: DataWrapper) -> Array:
 
         # https://github.com/pylint-dev/pylint/issues/3893
         # pylint: disable=unexpected-keyword-arg
-        return DataWrapper(
+        return DataWrapper(  # type: ignore[call-arg]
```
inducer, line 193: I actually don't understand what mypy is complaining about here. `shape` and `data` are defined as dataclass attributes?

alexfikl: Yeah, I really don't know either :\ I've stared a bit at the pytato code, but it's obviously fine. It's doubly surprising since those two attributes are actually in the class itself, but in a parent..

Definitely the same thing that's confusing pylint as well though.

alexfikl: For what it's worth, doing a `reveal_type(DataWrapper)` says
```
arraycontext/impl/pytato/utils.py:193: note: Revealed type is "def (*, axes: builtins.tuple[pytato.array.Axis, ...], tags: builtins.frozenset[pytools.tag.Tag], non_equality_tags: builtins.frozenset[pytools.tag.Tag] =) -> pytato.array.DataWrapper"
```
so for some reason it really doesn't think those arguments are there..

alexfikl: Possibly related to this: https://github.com/python/mypy/issues/17375

inducer: Good find, thanks. Added that as context in a comment so we don't lose it.

PR #290, opened 2024-11-27
Better types 2024

inducer:


arraycontext/container/__init__.py
```diff
@@ -145,6 +147,29 @@ class ArrayContainer(Protocol):
     # that are container-typed.
 
 
+class ArithArrayContainer(ArrayContainer, Protocol):
```
alexfikl, line 150: Does this need to be a subclass of `ArrayContainer`? Not sure how subclassing protocols is supposed to work..

arraycontext/container/dataclass.py
```diff
@@ -104,33 +114,46 @@ def is_array_field(f: Field) -> bool:
...
+        field_type_or_str = field.type
+        if isinstance(field_type_or_str, str):
+            if cls_ann is None:
+                cls_ann = get_annotations(cls, eval_str=True)
+            field_type = cls_ann[field.name]
+        else:
+            field_type = field_type_or_str
+        (array_fields
+            if is_array_field(field, field_type)
+            else non_array_fields).append(field)
```
alexfikl, line 144: This is a bit ugly.. maybe a normal if+else?

arraycontext/container/dataclass.py
```diff
@@ -104,33 +114,46 @@ def is_array_field(f: Field) -> bool:
...
-    from pytools import partition
-    array_fields, non_array_fields = partition(is_array_field, fields(cls))
+                        f"'{field_type!r}'")
+
+        return is_array_type(field_type)
+
+    array_fields: list[Field] = []
+    non_array_fields: list[Field] = []
+    from inspect import get_annotations
+    cls_ann: Mapping[str, type] | None = None
```
alexfikl, line 133:
```suggestion
    from inspect import get_annotations
    
    array_fields: list[Field] = []
    non_array_fields: list[Field] = []
    cls_ann: Mapping[str, type] | None = None
```
? It was confusing to see the import in the middle there :D

arraycontext/context.py
```diff
@@ -315,6 +360,14 @@ def from_numpy(self,
             intact.
         """
 
+    @overload
+    def to_numpy(self, array: np.ndarray) -> Array:
```
alexfikl, line 364:
```suggestion
    def to_numpy(self, array: Array) -> np.ndarray:
```
?

arraycontext/impl/pytato/utils.py
```diff
@@ -160,10 +163,7 @@ def map_data_wrapper(self, expr: DataWrapper) -> Array:
 
         # https://github.com/pylint-dev/pylint/issues/3893
         # pylint: disable=unexpected-keyword-arg
-        # type-ignore: discussed at
-        # https://github.com/inducer/arraycontext/pull/289#discussion_r1855523967
-        # possibly related: https://github.com/python/mypy/issues/17375
-        return DataWrapper(  # type: ignore[call-arg]
+        return DataWrapper(
```
alexfikl, line 166: This works now? What happened?

test/test_arraycontext.py
```diff
@@ -114,147 +113,10 @@ def _acf():
 # }}}
 
 
-# {{{ stand-in DOFArray implementation
-
-@with_container_arithmetic(
```
alexfikl, line 119: This doesn't need to move anymore, right? Since it doesn't have any string annotations.

I'm in favor of the little `testlib` regardless, so just curious :grin:

test/test_utils.py
```diff
@@ -49,29 +51,15 @@ def test_pt_actx_key_stringification_uniqueness():
...
-
-    # }}}
-
     # {{{ optional fields
 
     @dataclass
     class ArrayContainerWithOptional:
         x: np.ndarray
-        y: Optional[np.ndarray]
+        y: np.ndarray | None
```
alexfikl, line 62: This was intentionally left as `Optional[...]` (same for the `Tuple` and `Union` below). E.g. `Tuple[...]` and `tuple[...]` have different base classes, so `dataclass_container` needs to check for both.

test/test_utils.py
```diff
@@ -49,29 +51,15 @@ def test_pt_actx_key_stringification_uniqueness():
...
-    @dataclass
-    class ArrayContainerWithStringTypes:
-        x: np.ndarray
-        y: "np.ndarray"
-
-    with pytest.raises(TypeError, match="String annotation on field 'y'"):
-        # NOTE: cannot have string annotations in container
-        dataclass_array_container(ArrayContainerWithStringTypes)
-
-    # }}}
```
alexfikl, line 67: Maybe use some container from `testlib` to check this case actually works? I'm assuming those all have string fields due to `from __future__ import annotations`.

alexfikl (commented):
Left a few nitpicks, but overall adding some arithmetic to the types sounds good to me. 

Maybe it's worth even making the default `*T` versions be arithmetic? I can't think of a case where we'd want to be more general..

test/test_utils.py
alexfikl: Those are already tested, nevermind.. :facepalm:

arraycontext/container/__init__.py
inducer: - `ArrayContainer` is an empty protocol, so functionally the inheritance is a no-op, I think.
- An array container with arithmetic *is* an array container, so the inheritance "says the right thing" to me.

IDK. If you feel strongly, I'd be happy to remove it.

arraycontext/impl/pytato/utils.py
inducer: I don't understand it TBH. A few lines down, the symmetric type-ignore is still required. It had something to do with the better `{to,from}_numpy` types. So my guess is that the type of `new_dw` was off. (Which is actually misnamed a bit, it's not a `DataWrapper`---I'll rename it.)

test/test_arraycontext.py
inducer: If it stood on its own, yes, it would not need to move. But there are things in `testlib` that depend on it, and it's not importable here. It also just seems happier there.

test/test_utils.py
inducer: Fixed!

arraycontext/container/dataclass.py
inducer: Fixed!

arraycontext/container/dataclass.py
inducer: Fixed!

arraycontext/context.py
inducer: Whoops! Fixed.

inducer:
Thanks for reviewing!

arraycontext/container/__init__.py
alexfikl: No, that's perfectly fine!

arraycontext/impl/pytato/utils.py
alexfikl: Hm.. very weird. It was complaining about the attributes to the constructor, not `new_dw.data`.. Oh well, one less ignore :grin:

test/test_arraycontext.py
alexfikl: Ah, right, missed that.

PR #293, opened 2024-12-13
Fix RUF052 errors

alexfikl:
This fixes some genuine looking RUF052 errors where the function arguments just have a `_name` because they were meant not to shadow the outer scope.

### DIFF
--- a/arraycontext/loopy.py
+++ b/arraycontext/loopy.py
@@ -83,12 +83,14 @@ def get_default_entrypoint(t_unit):
 def _get_scalar_func_loopy_program(actx, c_name, nargs, naxes):
     @memoize_in(actx, _get_scalar_func_loopy_program)
     def get(c_name, nargs, naxes):
-        from pymbolic import var
+        from pymbolic.primitives import Subscript, Variable
 
         var_names = [f"i{i}" for i in range(naxes)]
         size_names = [f"n{i}" for i in range(naxes)]
-        subscript = tuple(var(vname) for vname in var_names)
+        subscript = tuple(Variable(vname) for vname in var_names)
+
         from islpy import make_zero_and_vars
+
         v = make_zero_and_vars(var_names, params=size_names)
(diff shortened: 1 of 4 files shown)


arraycontext/loopy.py
```diff
@@ -98,22 +100,22 @@ def get(c_name, nargs, naxes):
 
         import loopy as lp
 
-        from .loopy import make_loopy_program
         from arraycontext.transform_metadata import ElementwiseMapKernelTag
+
+        def sub(name: str) -> Variable | Subscript:
+            return Variable(name)[subscript] if subscript else Variable(name)
```
alexfikl, line 106: This is meant to silence the deprecation in `pymbolic` about an empty subscript. I went with keeping the same behavior here, i.e. returning the expression itself if the subscript is empty. Let me know if it looks ok!

alexfikl:
Hm, looks like the pylint bug is still not fixed :(

inducer:
> Hm, looks like the pylint bug is still not fixed :(

The corresponding issue is not an encouraging read: https://github.com/pylint-dev/pylint/issues/10112.

I think I may just disable pylint in ci-support if the Python version is 3.13.

inducer:
https://github.com/inducer/ci-support/commit/bf5975c3a25eb551567c9f799c49766801281f14 and https://github.com/inducer/ci-support/commit/1ddf2c37d188e571de4d7ebe5f7f4d02a02b2077.

alexfikl:
> > Hm, looks like the pylint bug is still not fixed :(
> 
> The corresponding issue is not an encouraging read: [pylint-dev/pylint#10112](https://github.com/pylint-dev/pylint/issues/10112).
> 
> I think I may just disable pylint in ci-support if the Python version is 3.13.

Not great, but that sounds fine by me. They'll probably get around to fixing it soon-ish.

inducer:
Thx!

PR #294, opened 2024-12-17
Refactor evaluating string fields

alexfikl:
I was updating https://github.com/inducer/boxtree/pull/56 and noticed that it also needs to handle string annotations, but with a slightly different `is_array_field` predicate. 

This refactors the `get_annotations` logic into a separate function so that it can be reused.

### DIFF
--- a/arraycontext/container/dataclass.py
+++ b/arraycontext/container/dataclass.py
@@ -32,13 +32,21 @@ THE SOFTWARE.
 """
 
 from collections.abc import Mapping, Sequence
-from dataclasses import Field, fields, is_dataclass
-from typing import Union, get_args, get_origin
+from dataclasses import fields, is_dataclass
+from typing import NamedTuple, Union, get_args, get_origin
 
 from arraycontext.container import is_array_container_type
 
 
 # {{{ dataclass containers
+
+class _Field(NamedTuple):
+    """Small lookalike for :class:`dataclasses.Field`."""
+
+    init: bool
+    name: str
+    type: type
+
 
 def is_array_type(tp: type) -> bool:
     from arraycontext import Array
@@ -73,7 +81,9 @@ def dataclass_array_container(cls: type) -> type:
 
     assert is_dataclass(cls)
 
-    def is_array_field(f: Field, field_type: type) -> bool:
+    def is_array_field(f: _Field) -> bool:
+        field_type = f.type
+
         # NOTE: unions of array containers are treated separately to handle
         # unions of only array containers, e.g. `Union[np.ndarray, Array]`, as
         # they can work seamlessly with arithmetic and traversal.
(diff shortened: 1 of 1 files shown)


arraycontext/container/dataclass.py
```diff
@@ -127,36 +136,52 @@ def is_array_field(f: Field, field_type: type) -> bool:
...
+    array_fields, non_array_fields = partition(is_array_field, array_fields)
+
+    if not array_fields:
+        raise ValueError(f"'{cls}' must have fields with array container type "
+                "in order to use the 'dataclass_array_container' decorator")
+
+    return _inject_dataclass_serialization(cls, array_fields, non_array_fields)
+
+
+def _get_annotated_fields(cls: type) -> Sequence[_Field]:
```
alexfikl, line 151: This could be a `_partition_fields(cls, predicate)` instead?

arraycontext/container/dataclass.py
```diff
@@ -32,14 +32,22 @@
...
-from typing import Union, get_args, get_origin
+from dataclasses import fields, is_dataclass
+from typing import NamedTuple, Union, get_args, get_origin
 
 from arraycontext.container import is_array_container_type
 
 
 # {{{ dataclass containers
 
+class _Field(NamedTuple):
```
inducer, line 43: Any reason to make this a `NamedTuple` and not a dataclass? I'm not even sure why/how to decide between the two.

arraycontext/container/dataclass.py
```diff
@@ -96,12 +106,11 @@ def is_array_field(f: Field, field_type: type) -> bool:
...
 
-        if isinstance(field_type, str):
-            raise TypeError(
-                f"String annotation on field '{f.name}' not supported. "
-                "(this may be due to 'from __future__ import annotations')")
-
         if __debug__:
+            # NOTE: this should never happen due to using `inspect.get_annotations`
+            if isinstance(field_type, str):
+                raise TypeError(f"String annotation on field '{f.name}' not supported.")
```
inducer, line 112:
```suggestion
                raise AssertionError(f"Unexpected string annotation on field '{f.name}'.")
```

inducer, line 112:
```suggestion
        assert not isinstance(field_type, str)
```
?

inducer (commented):
Thanks! Just a few minor nits/questions, but generally LGTM!

arraycontext/container/dataclass.py
alexfikl: Hm, not really.. I tend to use them for smaller private things, where I don't need default values, special inits, inheritance, etc.

I would expect them to be a bit faster and lighter if you don't need the extra features, but haven't really checked :\ Some more informed points: https://stackoverflow.com/a/51673969

arraycontext/container/dataclass.py
alexfikl: Yeah, went with that instead: https://github.com/inducer/arraycontext/compare/c44687e7444bd03b4b09c38bc19ea4125727325c..f705a8ad0bedc57175942f91dc07f08bdcc48557

alexfikl: Went with a simple assert in https://github.com/inducer/arraycontext/compare/c44687e7444bd03b4b09c38bc19ea4125727325c..f705a8ad0bedc57175942f91dc07f08bdcc48557.

arraycontext/container/dataclass.py
inducer: I'm OK with the way it is.

PR #299, opened 2025-01-08
Add an implementation of np.vdot to PytatoPyOpenCLArrayContext

### LINKED ISSUE #295: PytatoPyOpenCLArrayContext should support np.vdot() or something similar
I think the difficult part is that one would need to look at the data within the array like objects to determine if they are real or complex. 

However, if an array is tagged to hold complex data at the beginning then we can propagate the tag through the DAG keeping track of whether the data is still complex. Then, when we reach this new function, `vdot`, we can use the appropriate code.

Another way to do it would be to always take the complex conjugate even if the data is real-valued.

cc @inducer

nkoskelo:
closes #295

### DIFF
--- a/arraycontext/impl/pytato/fake_numpy.py
+++ b/arraycontext/impl/pytato/fake_numpy.py
@@ -239,4 +239,10 @@ class PytatoFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
     def absolute(self, a):
         return self.abs(a)
 
+    def vdot(self, a: Array, b: Array, order_a: str = "C", order_b: str = "C"):
+
+        flat_a = self.ravel(a, order_a)
+        flat_b = self.ravel(b, order_b)
+
+        return rec_multimap_array_container(pt.vdot, flat_a, flat_b)
     # }}}
--- a/arraycontext/impl/pytato/fake_numpy.py
+++ b/arraycontext/impl/pytato/fake_numpy.py
@@ -239,10 +239,7 @@ class PytatoFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
     def absolute(self, a):
         return self.abs(a)
 
-    def vdot(self, a: Array, b: Array, order_a: str = "C", order_b: str = "C"):
+    def vdot(self, a: Array, b: Array):
 
-        flat_a = self.ravel(a, order_a)
-        flat_b = self.ravel(b, order_b)
-
-        return rec_multimap_array_container(pt.vdot, flat_a, flat_b)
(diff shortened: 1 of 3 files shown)


test/test_arraycontext.py
```diff
@@ -271,9 +271,52 @@ def evaluate(np_, *args_):
...
+            ("conj", 1, np.complex128),
+            ("vdot", 2, np.float64),
+            ("vdot", 2, np.complex128),
+            ("abs", 1, np.float64),
+            ("abs", 1, np.complex128),
+            ("sum", 1, np.float64),
+            ("sum", 1, np.complex64),
+            ("isnan", 1, np.float64),
+            ])
+def test_array_context_np_workalike_with_scalars(actx_factory, sym_name, n_args, dtype):
```
inducer, line 302: This looks like largely a copy of the above test. Find a way to avoid the duplication.

arraycontext/impl/pytato/fake_numpy.py
```diff
@@ -239,4 +239,10 @@ def amin(self, a, axis=None):
     def absolute(self, a):
         return self.abs(a)
 
+    def vdot(self, a: Array, b: Array, order_a: str = "C", order_b: str = "C"):
+
+        flat_a = self.ravel(a, order_a)
+        flat_b = self.ravel(b, order_b)
+
+        return rec_multimap_array_container(pt.vdot, flat_a, flat_b)
```
inducer, line 247: Why `ravel` before calling pytato's `vdot`?

nkoskelo: The reshape option within pytato's ```vdot``` assumes that you have an array aranged in "C" order. However, the arraycontext's ```ravel``` option allows the user to specify the ordering.

test/test_arraycontext.py
nkoskelo: I split the old test function into two. This way we see if operation passes for Arrays and ArrayContainers as input. Previously, ```vdot``` and other functions that are not defined for scalar inputs would result in a skipped test even if it passed the other asserts within the test.

arraycontext/impl/pytato/fake_numpy.py
inducer: [numpy](https://numpy.org/doc/stable/reference/generated/numpy.vdot.html) does not have that feature. We should not add it, to avoid possible future incompatibility.

test/test_arraycontext.py
inducer: Can you just tweak the logic in the existing test, instead of introducing duplication?

PR #300, opened 2025-01-10
Use Optional and Tuple to test decorator

alexfikl:
This adds back the `Optional[Array]` that was removed in #299 to fix the ruff warnings.

These are mostly used to test the functionality of the `dataclass_array_container` decorator. Do we want to remove support for old-style `Optional` / `Tuple` / `List` / etc from there instead?

cc @nkoskelo


inducer:
Thanks for catching that. I had missed that during the review. I think we can keep the support for the old-style annotations at least for a bit.

PR #305, opened 2025-03-18
upgrade 'unevaluated array as argument' warning to error

matthiasdiener:
_Please squash_

### DIFF
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -620,7 +620,7 @@ class CompiledPyOpenCLFunctionReturningArrayContainer(CompiledFunction):
        type of the callable.
     """
     actx: PytatoPyOpenCLArrayContext
-    pytato_program: pt.target.BoundProgram
+    pytato_program: pt.target.loopy.BoundPyOpenCLExecutable
     input_id_to_name_in_program: Mapping[tuple[Hashable, ...], str]
     output_id_to_name_in_program: Mapping[tuple[Hashable, ...], str]
     name_in_program_to_tags: Mapping[str, frozenset[Tag]]
@@ -665,7 +665,7 @@ class CompiledPyOpenCLFunctionReturningArray(CompiledFunction):
         Name of the output array in the program.
     """
     actx: PytatoPyOpenCLArrayContext
-    pytato_program: pt.target.BoundProgram
+    pytato_program: pt.target.loopy.BoundPyOpenCLExecutable
     input_id_to_name_in_program: Mapping[tuple[Hashable, ...], str]
     output_tags: frozenset[Tag]
     output_axes: tuple[pt.Axis, ...]
@@ -675,7 +675,7 @@ class CompiledPyOpenCLFunctionReturningArray(CompiledFunction):
         from .utils import get_cl_axes_from_pt_axes
         from arraycontext.impl.pyopencl.taggable_cl_array import to_tagged_cl_array
 
-        fn_name = self.pytato_program.program.name
+        fn_name = self.pytato_program.program.entrypoint
 
         input_kwargs_for_loopy = _args_to_device_buffers(
                 self.actx, self.input_id_to_name_in_program, arg_id_to_arg, fn_name)
--- a/arraycontext/impl/pytato/compile.py
+++ b/arraycontext/impl/pytato/compile.py
@@ -564,15 +564,6 @@ def _args_to_device_buffers(actx, input_id_to_name_in_program, arg_id_to_arg):
 
     return input_kwargs_for_loopy
 
-
-def _args_to_cl_buffers(actx, input_id_to_name_in_program, arg_id_to_arg):
-    from warnings import warn
-    warn("_args_to_cl_buffer has been renamed to"
-         " _args_to_device_buffers. This will be"
-         " an error in 2023.", DeprecationWarning, stacklevel=2)
-    return _args_to_device_buffers(actx, input_id_to_name_in_program,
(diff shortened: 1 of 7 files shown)


inducer:
Could you explain the context here? I'd imagine that this will be awfully noisy, for not much gain. Ultimately, one should fix these warnings until they're all gone.

matthiasdiener:
> Could you explain the context here? I'd imagine that this will be awfully noisy, for (IMO) not much gain. Ultimately, one should fix these warnings until they're all gone.

My thought here was that in contrast to most (almost all?) other warnings, this type of warning can have serious performance implications. I was surprised to see this warning appear in the shock1d driver, and they appeared at a point during the execution where *many* other warnings are displayed, and thus this kind of warnings tends to "disappear". I'll set this back to draft, I think we should probably show the name of the compiled function.

alexfikl:
> logger.warning does not deduplicate, in contrast to warnings.warn

You can set `warnings.filterwarnings("always")` to remove deduplication from warnings (can probably even be targeted at this particular warning). Would that not be the same?

> My thought here was that in contrast to most (almost all?) other warnings, this type of warning can have serious performance implications.

If that's the case, wouldn't it be better to disable all warnings on a "production run"?

matthiasdiener:
> > logger.warning does not deduplicate, in contrast to warnings.warn
> 
> You can set `warnings.filterwarnings("always")` to remove deduplication from warnings (can probably even be targeted at this particular warning). Would that not be the same?

TIL, thanks. The interface is a bit clumsy, but it seems to work.

> 
> > My thought here was that in contrast to most (almost all?) other warnings, this type of warning can have serious performance implications.
> 
> If that's the case, wouldn't it be better to disable all warnings on a "production run"?

Sorry, my text probably wasn't very clear here. What I meant is that the issue pointed out by the warnings could have serious implications.

alexfikl:
> TIL, thanks. The interface is a bit clumsy, but it seems to work.

Oops, sorry, I wasn't clear there.. I meant that you can just filter warnings in your top-level driver, not here. I agree that it looks very clumsy as is :(

> Sorry, my text probably wasn't very clear here. What I meant is that the issue pointed out by the warnings could have serious implications.

Ah, so the issue is that the warning is not loud enough? I would still argue that this is better done on the application side with `filterwarnings`, but don't feel very strongly about it..

arraycontext/impl/pytato/compile.py
```diff
@@ -550,13 +550,15 @@ def _args_to_device_buffers(actx, input_id_to_name_in_program, arg_id_to_arg):
...
-                    DeprecationWarning, stacklevel=4)
+            from warnings import catch_warnings, filterwarnings, warn
+            with catch_warnings():
+                filterwarnings("always")
+                warn(f"Argument array '{arg_id}' to the '{fn_name}' compiled function "
+                     "is unevaluated. Evaluating just-in-time, at "
+                     "considerable expense. This is deprecated and will stop "
+                     "working in 2023. To avoid this warning, force evaluation "
+                     "of all arguments via freeze/thaw.",
+                      stacklevel=4)
```
inducer, line 561: I think it'd be useful to create a special warning class for this:
```
class UnevaluatedCompiledFuncArgWarning(UserWarning):
    pass
```
(and then pass the appropriate argument to `warn`) This makes precise filtering from the application side possible. I don't know that I want to install a warning filter on the user's behalf. (We could, in the actx constructor, on an opt-out basis? But that still feels overbearing.)

Alternatively, we could just make this an error, given that it has been deprecated with a deadline that has passed.

inducer: I'm starting to like the "make it an error" idea, I think.

matthiasdiener: > I'm starting to like the "make it an error" idea, I think.

Same here :)

matthiasdiener: Done in [5535398](https://github.com/inducer/arraycontext/pull/305/commits/553539813f442f85cf6c92765f8854dc1c0b5b8e)

inducer:
https://github.com/inducer/grudge/pull/376 should address the grudge failures.

arraycontext/impl/pytato/compile.py
```diff
@@ -642,8 +631,10 @@ def __call__(self, arg_id_to_arg) -> ArrayContainer:
         from .utils import get_cl_axes_from_pt_axes
         from arraycontext.impl.pyopencl.taggable_cl_array import to_tagged_cl_array
 
+        fn_name = self.pytato_program.program.entrypoint
```
inducer, line 634: This and the `fn_name`s below appear to be inconsistent. Is there a specific reason for this?

inducer (commented):
LGTM, just one question, hopefully quick.

arraycontext/impl/pytato/compile.py
matthiasdiener: Thanks, this is hopefully fixed in b1c3a73. Part of the challenge here is that `CompiledPyOpenCLFunctionReturningArray` appears to be unused.

majosm:
FYI @inducer @matthiasdiener, this change appears to break the non-compiled time integrators in grudge/mirgecom. (Looks like grudge downstream CI here doesn't run the examples, so it missed that.)

inducer:
> This change appears to break the non-compiled time integrators in grudge/mirgecom.

As well it should! I'll get grudge fixed up.

inducer:
See https://github.com/inducer/grudge/pull/377 and https://github.com/inducer/arraycontext/pull/307.

PR #306, opened 2025-03-21
Specialize `freeze_thaw` for pytato

inducer:


arraycontext/impl/pytato/__init__.py
```diff
@@ -338,22 +342,14 @@ def _rec_map_container(
...
-                return func(tga.to_tagged_cl_array(ary))
             elif np.isscalar(ary):
                 if default_scalar is None:
                     return ary
                 else:
                     return np.array(ary).dtype.type(default_scalar)
             else:
                 raise TypeError(
-                    f"{type(self).__name__}.{func.__name__[1:]} invoked with "
+                    f"{func.__qualname__}.{func.__name__} invoked with "
```
matthiasdiener, line 352: Isn't just __qualname__ enough?

```suggestion
                    f"{func.__qualname__} invoked with "
```

inducer: You're right! Fixed.

matthiasdiener (approved):

PR #307, opened 2025-03-27
Add `bcast` and `bcast_until_actx_array functions`

inducer:
I'm mostly seeking early feedback on this. It still needs, at least:

- [ ] Documentation
- [ ] Tests
- [ ] Fixed types

This could help fill the hole that's left by the plan to make implicit broadcasting of data class containers across actx array types illegal. See https://github.com/inducer/grudge/pull/377 for an example use, specifically `shortcuts.py`.

Closes #280. (Would supersede the cursed `Bcast` objects that @alexfikl hated---and I have to agree, it is just nicer.)

### DIFF
--- a/arraycontext/container/traversal.py
+++ b/arraycontext/container/traversal.py
@@ -78,6 +78,7 @@ from warnings import warn
 import numpy as np
 
 from arraycontext.container import (
+    ArithArrayContainer,
     ArrayContainer,
     NotAnArrayContainerError,
     SerializationKey,
@@ -88,6 +89,7 @@ from arraycontext.container import (
 from arraycontext.context import (
     Array,
     ArrayContext,
+    ArrayOrArithContainer,
     ArrayOrContainer,
     ArrayOrContainerOrScalar,
     ArrayOrContainerT,
@@ -986,5 +988,80 @@ def outer(a: Any, b: Any) -> Any:
         return multimap_array_container(lambda x, y: outer(x, y), a, b)
 
 # }}}
+
+
+# {{{
+
+def bcast_left(
+            op: Callable[[ArrayOrArithContainer, ArrayOrArithContainer],
+                    ArrayOrArithContainer],
+            left: ArrayOrArithContainer,
+            right: ArithArrayContainer,
+        ) -> ArrayOrArithContainer:
+    try:
+        serialized = serialize_container(right)
+    except NotAnArrayContainerError:
+        return op(left, right)
+
+    return deserialize_container(right, [
+        (k, op(left, right_v)) for k, right_v in serialized])
+
+
+def bcast_right(
+            op: Callable[[ArrayOrArithContainer, ArrayOrArithContainer],
+                    ArrayOrArithContainer],
+            left: ArrayOrArithContainer,
+            right: ArithArrayContainer,
+        ) -> ArrayOrArithContainer:
+    try:
+        serialized = serialize_container(left)
+    except NotAnArrayContainerError:
+        return op(left, right)
+
+    return deserialize_container(right, [
+        (k, op(left_v, right)) for k, left_v in serialized])
+
+
+def bcast_left_until_actx_array(
+            actx: ArrayContext,
+            op: Callable[[ArrayOrArithContainer, ArrayOrArithContainer],
+                    ArrayOrArithContainer],
+            left: ArrayOrArithContainer,
+            right: ArithArrayContainer,
+        ) -> ArrayOrArithContainer:
+    try:
+        serialized = serialize_container(right)
+    except NotAnArrayContainerError:
+        return op(left, right)
+
+    return deserialize_container(right, [
+        (k, op(left, right_v)
+            if isinstance(right_v, actx.array_types) else
+            bcast_left_until_actx_array(actx, op, left, right_v)
+        )
+        for k, right_v in serialized])
+
+
+def bcast_right_until_actx_array(
+            actx: ArrayContext,
+            op: Callable[[ArrayOrArithContainer, ArrayOrArithContainer],
+                    ArrayOrArithContainer],
+            left: ArrayOrArithContainer,
+            right: ArithArrayContainer,
+        ) -> ArrayOrArithContainer:
+    try:
+        serialized = serialize_container(left)
+    except NotAnArrayContainerError:
+        return op(left, right)
+
+    return deserialize_container(right, [
+        (k, op(left_v, right)
+            if isinstance(left_v, actx.array_types) else
+            bcast_right_until_actx_array(actx, op, left_v, right)
+        )
+        for k, left_v in serialized])
+
+# }}}
+
 
 # vim: foldmethod=marker
--- a/arraycontext/__init__.py
+++ b/arraycontext/__init__.py
@@ -50,6 +50,10 @@ from .container.arithmetic import (
 )
 from .container.dataclass import dataclass_array_container
 from .container.traversal import (
+    bcast_left,
+    bcast_left_until_actx_array,
+    bcast_right,
+    bcast_right_until_actx_array,
     flat_size_and_dtype,
     flatten,
     freeze,
@@ -129,6 +133,10 @@ __all__ = (
     "ScalarLike",
     "SerializationKey",
     "SerializedContainer",
+    "bcast_left",
+    "bcast_left_until_actx_array",
+    "bcast_right",
+    "bcast_right_until_actx_array",
     "dataclass_array_container",
     "deserialize_container",
     "flat_size_and_dtype",
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -413,10 +413,8 @@ def with_container_arithmetic(
                 warn(
                     f"Broadcasting array context array types across {cls} "
                     "has been implicitly "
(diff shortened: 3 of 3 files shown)


arraycontext/container/traversal.py
```diff
@@ -987,4 +989,79 @@ def treat_as_scalar(x: Any) -> bool:
...
+    try:
+        serialized = serialize_container(left)
+    except NotAnArrayContainerError:
+        return op(left, right)
+
+    return deserialize_container(right, [
+        (k, op(left_v, right)) for k, left_v in serialized])
+
+
+def bcast_left_until_actx_array(
```
inducer, line 1025: Naming? (May not matter as much, as it's convenient to make local aliases, as in the grudge PR.)

inducer, line 1025: Argument order? (Here and above.) I initially had the operator in the middle but went with "first" to allow convenient use of `partial`.

arraycontext/container/traversal.py
```diff
@@ -987,4 +989,79 @@ def treat_as_scalar(x: Any) -> bool:
 
 # }}}
 
+
+# {{{
```
inducer, line 993:
```suggestion
# {{{ bcast
```

inducer, line 993: Should these be here or in `arithmetic`?

inducer (commented):
Some specific questions for the review.

alexfikl (approved):
Took a quick look and it seems fine to me. 

The only issue would be if it's too verbose for more complex usage? The one in grudge looks nice enough to me :\

arraycontext/container/traversal.py
alexfikl: Maybe `rec_bcast_left`? Mostly to match some of the other traversal functions. 

I agree that local aliases are probably needed, so this should be nice and verbose.. i.e. I'm quite fine with the name.

alexfikl: I like the functional feel of it too :+1:

arraycontext/container/traversal.py
alexfikl: I imagine they'll be pretty exclusively imported from the top level.. so, meh? `arithmetic` is probably fine just containing the dunder methods.

inducer:
Thanks for taking a look!

> The one in grudge looks nice enough to me :\

Same.

> The only issue would be if it's too verbose for more complex usage?

Good question. Without question, the prior state in grudge is nicer to look at.

Before:
```
residual = a*residual + h*rhs_val
y = y + b * residual
```

After:
```
residual = bcmul(a, residual) + bcmul(h, rhs_val)
y = y + bcmul(b, residual)
```
On the one hand, it's clearly more verbose. On the other, the verbosity scales linearly with expression size, so maybe not too bad? We could also just defer this issue to when it comes up. One attractive feature of these functions is that they're easy to replace: deprecate, grep, done. So I anticipate this to be low regret, even if it turns out to not be the right answer long term.

If we wanted to preserve the operatory-y feel, we need a way to let the the operator know broadcast rules (given that they depend on the array context), and we are (I am) hoping to no longer have array context handles in array containers. The only approach I can think of there is the `Bcast` solution (https://github.com/inducer/arraycontext/pull/280) for reference:
```
residual = bcast(a)*residual + bcast(h)*rhs_val
y = y + bcast(b) * residual
```
(That's assuming a local alias `bcast` has been created.) I'm not advocating for it. It's just about the same amount of visual clutter, and it requires special machinery inside container arithmetic, unlike the function-based approach here.

alexfikl:
> The only approach I can think of there is the `Bcast` solution (#280) for reference:
> 
> ```
> residual = bcast(a)*residual + bcast(h)*rhs_val
> y = y + bcast(b) * residual
> ```
> (That's assuming a local alias bcast has been created.) I'm not advocating for it. It's just about the same amount of visual clutter, and it requires special machinery inside container arithmetic, unlike the function-based approach here.

That's a good point.. I forgot that version was quite verbose-y as well. Then yeah, from my side this looks very nice and very low on the black-box magic compared to the previous version :grin:

arraycontext/container/traversal.py
majosm: Agree with adding `rec_`, and I would suggest maybe adding something else to disambiguate what's happening to the "left". (Picturing myself looking at this again in 6 months and not being sure if it means "broadcast the left across" or "broadcast across the left".) `rec_bcast_left_operand_across_actx_arrays` or something (probably too verbose, maybe you can think of something better).

arraycontext/container/traversal.py
```diff
@@ -987,4 +989,79 @@ def treat_as_scalar(x: Any) -> bool:
...
+        (k, op(left_v, right)) for k, left_v in serialized])
+
+
+def bcast_left_until_actx_array(
+            actx: ArrayContext,
+            op: Callable[[ArrayOrArithContainer, ArrayOrArithContainer],
+                    ArrayOrArithContainer],
+            left: ArrayOrArithContainer,
+            right: ArithArrayContainer,
+        ) -> ArrayOrArithContainer:
```
majosm, line 1031: Would it make sense to have the interface for these be more like the container traversal functions, e.g.:
```python
def rec_bcast_left(op, left, right, leaf_cls: type | None = None):
    ...
```
and then
```python
mul = partial(rec_bcast_left, operator.mul, leaf_cls=actx.array_types)
```
?

majosm:
> Thanks for taking a look!
> 
> > The one in grudge looks nice enough to me :\
> 
> Same.
> 
> > The only issue would be if it's too verbose for more complex usage?
> 
> Good question. Without question, the prior state in grudge is nicer to look at.
> 
> Before:
> 
> ```
> residual = a*residual + h*rhs_val
> y = y + b * residual
> ```
> 
> After:
> 
> ```
> residual = bcmul(a, residual) + bcmul(h, rhs_val)
> y = y + bcmul(b, residual)
> ```
> 
> On the one hand, it's clearly more verbose. On the other, the verbosity scales linearly with expression size, so maybe not too bad? We could also just defer this issue to when it comes up. One attractive feature of these functions is that they're easy to replace: deprecate, grep, done. So I anticipate this to be low regret, even if it turns out to not be the right answer long term.
> 
> If we wanted to preserve the operatory-y feel, we need a way to let the the operator know broadcast rules (given that they depend on the array context), and we are (I am) hoping to no longer have array context handles in array containers. The only approach I can think of there is the `Bcast` solution (#280) for reference:
> 
> ```
> residual = bcast(a)*residual + bcast(h)*rhs_val
> y = y + bcast(b) * residual
> ```
> 
> (That's assuming a local alias `bcast` has been created.) I'm not advocating for it. It's just about the same amount of visual clutter, and it requires special machinery inside container arithmetic, unlike the function-based approach here.

TBH from a syntax point of view I think I like the `bcast(a)` version better, as it avoids the "left"/"right" naming ambiguity that I mentioned above, and it seems a little easier to follow with the original operators still being there. (IIRC my main objection to #280 was the `1`/`2`/`3`/`NLevels` stuff.) If there aren't any major complications with doing it that way vs. the `mul` version, I think I would prefer that.

majosm:
Does the `bcast(a)` version necessitate doing the `_rewrap` stuff from #280? Or is there a way to do it that's more like the current implementation here?

inducer:
> Does the `bcast(a)` version necessitate doing the `_rewrap` stuff from #280?


Good question.

If we wrap the *thing being broadcast* in `Bcast` (as in #280), then I don't see how one can get by without the `rewrap` nonsense.

However, if we wrap *the container being broadcast over* in `Bcast`, conceivably we can get by with just overloaded operators on the `Bcast` object itself. I think that would considerably simplify the machinery (to just some overloaded operators). @alexfikl How much would you loathe that?

majosm:
> > Does the `bcast(a)` version necessitate doing the `_rewrap` stuff from #280?
> 
> Good question.
> 
> If we wrap the _thing being broadcast_ in `Bcast` (as in #280), then I don't see how one can get by without the `rewrap` nonsense.
> 
> However, if we wrap _the container being broadcast over_ in `Bcast`, conceivably we can get by with just overloaded operators on the `Bcast` object itself. I think that would considerably simplify the machinery (to just some overloaded operators). @alexfikl How much would you loathe that?

I can't speak for Alex, but I would object to that. 🙂 I would be confused by that syntax.

I'm probably missing something obvious, but why wouldn't we be able to do something along the lines of:

```python
class Bcast:
    def __init__(self, array):
      self.array = array

    def _binary_op(self, op, right):
        # Need a reverse case too, but I'm lazy

        try:
            serialized = serialize_container(right)
        except NotAnArrayContainerError:
            return op(self.array, right)

        return deserialize_container(right, [
            (k, op(self.array, right_v)
                if isinstance(right_v, actx.array_types) else
                self._binary_op(op, right_v)
            )
            for k, right_v in serialized])

    __mul__ = partialmethod(_binary_op, operator.mul)
```
?

inducer:
To add: I'm not hating that as a possible direction. It shouldn't need any weird "magic". @alexfikl?

alexfikl:
> To add: I'm not hating that as a possible direction. It shouldn't need any weird "magic". @alexfikl?

Agreed! That looks like it should be pretty easy to understand with some of the same greppability. I'll mull it over, but can't think of any downsides at the moment :-?

PR #312, opened 2025-04-04
Fix ruff 0.11.4 error

matthiasdiener:
Imho, not really clearer than before...

### DIFF
--- a/arraycontext/impl/pyopencl/taggable_cl_array.py
+++ b/arraycontext/impl/pyopencl/taggable_cl_array.py
@@ -119,9 +119,9 @@ class TaggableCLArray(cla.Array, Taggable):
         """
         Returns a copy of *self* with *iaxis*-th axis tagged with *tags*.
         """
-        new_axes = ((*self.axes[:iaxis],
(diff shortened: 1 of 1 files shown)


arraycontext/impl/pyopencl/taggable_cl_array.py
```diff
@@ -119,9 +119,9 @@ def with_tagged_axis(self, iaxis: int,
         """
         Returns a copy of *self* with *iaxis*-th axis tagged with *tags*.
         """
-        new_axes = (self.axes[:iaxis]
-                    + (self.axes[iaxis].tagged(tags),)
-                    + self.axes[iaxis+1:])
+        new_axes = ((*self.axes[:iaxis],
+                     self.axes[iaxis].tagged(tags),
+                     *self.axes[iaxis + 1:]))
```
alexfikl, line 124:
```suggestion
        new_axes = (*self.axes[:iaxis],
                    self.axes[iaxis].tagged(tags),
                    *self.axes[iaxis + 1:])
```
? Yeah, I think the main selling point of this is that it works with any iterable, which is not a big plus here..

PR #313, opened 2025-04-09
Attempt to make `BcastUntilActxArray` work for broadcastees that are containers

majosm:
In some mirgecom examples, `dt` can be a `DOFArray`. When using `_compiled_lsrk45_step` from grudge (which now uses `BcastUntilActxArray` for `h`, etc.), this causes `dt` to broadcast down into the array context arrays of the RHS container, leading to the creation of erroneous nested `DOFArray`s. So it looks like we need to tweak `BcastUntilActxArray` to handle the case of `broadcastee` being an array container.

I took a stab at doing so here. I don't know if this is the right approach, but my hunch is that we should let the container's defined broadcasting behavior take over in this case (mainly because I feel like, e.g., `bcast(Container()) + Container()` ought to be equivalent to `Container() + Container()` and not do any kind of broadcasting inside). Thoughts @inducer, @alexfikl?

### DIFF
--- a/arraycontext/container/arithmetic.py
+++ b/arraycontext/container/arithmetic.py
@@ -764,6 +764,10 @@ class BcastUntilActxArray:
                    ],
                    right: ArrayOrContainer
                ) -> ArrayOrContainer:
+        if not isinstance(self.broadcastee, self._stop_types):
+            # Defer to array container broadcast rules
+            return op(self.broadcastee, right)
+
         try:
             serialized = serialize_container(right)
         except NotAnArrayContainerError:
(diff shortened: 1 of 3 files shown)


inducer:
> In some mirgecom examples, `dt` can be a `DOFArray`.

Could you explain? What does it mean if the timestep is a `DOFArray`?

majosm:
> > In some mirgecom examples, `dt` can be a `DOFArray`.
> 
> Could you explain? What does it mean if the timestep is a `DOFArray`?

It's from Tulio's local-timestepping stuff. The `dt` value can vary per element.

PR #315, opened 2025-05-23
EagerJAXArrayContext: return mutable numpy array in to_numpy

matthiasdiener:
AFAICS, there is no way to make an immutable numpy array mutable again except for copying it.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -3120,30 +3120,6 @@
             {
                 "code": "reportUnknownVariableType",
                 "range": {
-                    "startColumn": 53,
-                    "endColumn": 62,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAny",
-                "range": {
-                    "startColumn": 23,
-                    "endColumn": 46,
-                    "lineCount": 1
-                }
-            },
-            {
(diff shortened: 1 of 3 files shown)


alexfikl:
I'm a bit out of the loop, so I have questions :grin:

1. Why is this required? To my knowledge, most of the code doesn't do in-place modifications (e.g. `DOFArray` in-place operations are very deprecated).
2. Haven't tested, but [jax.device_get](https://docs.jax.dev/en/latest/_autosummary/jax.device_get.html) docs mention copies being done. Is that not the case?
3. Would this be useful when using the CPU backend for jax? I image the GPU/TPU backends do a copy anyway?

alexfikl:
> There are in-place modifications in grudge, e.g. `grp_field` in [inducer/grudge@`193100a`/grudge/geometry/metrics.py#L567-L577](https://github.com/inducer/grudge/blob/193100abd566719b26001b85feb3d7d2171f5b8d/grudge/geometry/metrics.py#L567-L577)

I remember writing that code.. it should really be rewritten in a nicer way :(

But fair enough, I can see how that breaks. Are there any other places? (the tests in https://github.com/inducer/grudge/pull/380 seem to be passing nicely at the moment)

PR #325, opened 2025-07-01
Avoid duplication in `PytatoJAXArrayContext.freeze`

majosm:
Missed one. 😞


inducer:
> FYI, a test failed the first time I ran CI on this: https://github.com/inducer/arraycontext/actions/runs/16008044379/job/45159156740?pr=325. Seems unrelated to the changes here?

Thanks for flagging that. I ran this test a bunch of times locally and was unable to reproduce the failure; we'll have to leave this be for the time being.

PR #331, opened 2025-07-25
Remove dataclass from BaseFakeNumpyNamespace

alexfikl:
Not quite sure why this was made a dataclass in #322, but none of the subclasses were updated to be dataclasses too? This made some things not get picked up properly (as far as I can tell), like the updated attribute type
```python
class PyOpenCLFakeNumpyNamespace(LoopyBasedFakeNumpyNamespace):
    _array_context: PyOpenCLArrayContext
```

### DIFF
--- a/arraycontext/fake_numpy.py
+++ b/arraycontext/fake_numpy.py
@@ -29,7 +29,6 @@ THE SOFTWARE.
 
 import operator
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


inducer:
> This made some things not get picked up properly (as far as I can tell), like the updated attribute type

Could you explain? What didn't get picked up? By whom? To what end? And: I'm not sure I understand why it's a problem that some classes in an inheritance graph are data classes and others not. That seems fine, no?

inducer:
> EDIT: Adding a `np: PyOpenCLFakeNumpyNamespace` to `PyOpenCLArrayContext` seems to appease the type checkers, but `pylint` is still unhappy..

This seems like the way to go IMO. I'm not worried about pylint not "getting" it; I've already dropped it from CI.

PR #333, opened 2025-07-31
Support subclass fields in dataclass_array_container

alexfikl:
`get_annotations` only gets annotations from the current class, but not from any subclasses. This extends that by going through the MRO (in reverse) and gathering all of them up.

Context: `boxtree.Tree` inherits from `boxtree.TreeOfBoxes` and this wouldn't recognize the fields from `TreeOfBoxes`, which are still in the list returned by `fields(Tree)`.

### DIFF
--- a/arraycontext/container/dataclass.py
+++ b/arraycontext/container/dataclass.py
@@ -163,14 +163,16 @@ def _get_annotated_fields(cls: type) -> Sequence[_Field]:
     from inspect import get_annotations
 
     result = []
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! Added a test of sorts, plus some other coincidental typing stuff.

alexfikl:
> Thanks! Added a test of sorts, plus some other coincidental typing stuff.

Thanks for adding the test!

PR #335, opened 2025-08-26
Array(Protocol): rename __getitem__ parameter

matthiasdiener:
Context: Attempt to fix bpr in https://github.com/inducer/grudge/pull/380/

### DIFF
--- a/arraycontext/typing.py
+++ b/arraycontext/typing.py
@@ -153,7 +153,7 @@ class Array(Protocol):
     # If you'd  like to see how, try changing the Any to
     # AxisIndex = slice | int | "Array"
     # Index = AxisIndex |tuple[AxisIndex]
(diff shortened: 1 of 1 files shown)


arraycontext/typing.py
```diff
@@ -153,7 +153,7 @@ def dtype(self) -> np.dtype[Any]:
     # If you'd  like to see how, try changing the Any to
     # AxisIndex = slice | int | "Array"
     # Index = AxisIndex |tuple[AxisIndex]
-    def __getitem__(self, index: Any) -> Array:  # pyright: ignore[reportAny]
+    def __getitem__(self, key: Any) -> Array:  # pyright: ignore[reportAny]
```
inducer, line 156:
```suggestion
    def __getitem__(self, /, key: object) -> Array:
```
?

PR #357, opened 2026-05-10
PyOpenCLActx: fixes to np.where, actx.to_numpy

kaushikcfd:
Please refer to commit-by-commit description for the two fixes.

### DIFF
--- a/arraycontext/impl/pyopencl/__init__.py
+++ b/arraycontext/impl/pyopencl/__init__.py
@@ -37,6 +37,8 @@ from warnings import warn
 import numpy as np
 from typing_extensions import Self, override
 
+from pytools import memoize_method
+
 from arraycontext.container.traversal import (
     rec_map_array_container,
     rec_map_container,
@@ -62,7 +64,7 @@ from arraycontext.typing import (
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping
 
-    from numpy.typing import NDArray
+    from numpy.typing import DTypeLike, NDArray
 
     import loopy as lp
     import pyopencl as cl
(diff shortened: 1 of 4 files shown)


arraycontext/impl/pyopencl/__init__.py
```diff
@@ -263,12 +265,64 @@ def to_numpy(self, array: Array) -> np.ndarray:
     def to_numpy(self, array: ContainerOrScalarT) -> ContainerOrScalarT:
         ...
 
+    @memoize_method
+    def _get_to_numpy_contiguous_copy_kernel(
```
alexfikl:
```suggestion
    def _get_to_numpy_noncontiguous_copy_kernel(
```
? and in other places.

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -263,12 +265,64 @@ def to_numpy(self, array: Array) -> np.ndarray:
...
+        return t_unit
+
     @override
     def to_numpy(self,
                  array: ArrayOrContainerOrScalar
                  ) -> NumpyOrContainerOrScalar:
         def _to_numpy(ary):
-            return ary.get(queue=self.queue)
+            if ary.flags.forc:
+                # pyopenclsupports host transfers only for contiguous arrays.
```
alexfikl:
```suggestion
                # pyopencl supports host transfers only for contiguous arrays.
```

arraycontext/impl/pyopencl/fake_numpy.py
```diff
@@ -463,6 +463,31 @@ def where(self, criterion, then, else_):
         def where_inner(inner_crit, inner_then, inner_else):
             if isinstance(inner_crit, bool | np.bool_):
                 return inner_then if inner_crit else inner_else
+
+            # pyopencl's if_positive does not support then, else branches with
+            # unequal dtypes -> cast them to a common dtype.
```
alexfikl, line 473: Just to ask: worth adding this to `pyopencl` instead?

alexfikl (commented):
Looks good to me! Left a couple of nitpicks, but besides that:
* Where did these show up? For my own curiosity :grin:
* Add some small tests? Is the transpose test supposed to test the non-continuous copy?

arraycontext/impl/pyopencl/__init__.py
kaushikcfd: Done. Thanks!

arraycontext/impl/pyopencl/__init__.py
kaushikcfd: Done. thanks!

arraycontext/impl/pyopencl/fake_numpy.py
kaushikcfd: 1. I thought it was by design in pyopencl to keep the kernel launches transparent.
2. I wanted this with medium urgency couldn't wait for pyopencl's beauracracy for conda-forge, pypi to pick the new versions, etc etc.

kaushikcfd:
> Add some small tests? Is the transpose test supposed to test the non-continuous copy?

The linked regression exercises the code path. Pyopencl.Array.transpose is a no-op on the cl.Device simply creates a copy of the buffer with different strides.

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -263,12 +265,74 @@ def to_numpy(self, array: Array) -> np.ndarray:
     def to_numpy(self, array: ContainerOrScalarT) -> ContainerOrScalarT:
         ...
 
+    @memoize_method
+    def _get_to_numpy_noncontiguous_copy_kernel(
```
inducer, line 269: It seems this could be a `lp.make_copy_kernel` with some transforms? (If it's still needed, see below.)

arraycontext/impl/pyopencl/__init__.py
```diff
@@ -263,12 +265,74 @@ def to_numpy(self, array: Array) -> np.ndarray:
...
     def to_numpy(self,
                  array: ArrayOrContainerOrScalar
                  ) -> NumpyOrContainerOrScalar:
         def _to_numpy(ary):
-            return ary.get(queue=self.queue)
+            if ary.flags.forc:
+                # pyopencl supports host transfers only for contiguous arrays.
+                return ary.get(queue=self.queue)
+
+            result = self.call_loopy(
```
inducer, line 307: This doesn't make a great deal of sense to me.

- If host-to-GPU transfers fail, then this will fail, too, because the first thing that this `call_loopy` will do is attempt to transfer the array.
- Why not simply copy-to-contiguous on the host and then go from there? (And issue a warning for the hidden cost?)
- IMO the nicest solution would be to make use of OpenCL's strided copy primitive. There was [some code for that](https://github.com/inducer/pyopencl/pull/54/changes) at a point, but it wasn't tested and (unnecessarily) only worked in the `get` direction.

kaushikcfd: > the first thing that this call_loopy will do is attempt to transfer the array.
> Why not simply copy-to-contiguous on the host and then go from there?

I don't think I get these. Over here, we are doing `input_ary.copy(order="C")` through this loopy kernel on the _device_ before calling ".get".

> IMO the nicest solution would be to make use of OpenCL's strided copy primitive. There was [some code for that](https://github.com/inducer/pyopencl/pull/54/changes) at a point, but it wasn't tested and (unnecessarily) only worked in the get direction.

Agreed. I had seen that, but even if it were merged, it would only work for n-d arrays for n < 4 -- so we will still need this code for the general case.

arraycontext/impl/pyopencl/__init__.py
kaushikcfd: Done!

arraycontext/impl/pyopencl/__init__.py
inducer: > I don't think I get these.

I was confused, never mind.

inducer:
Thx!

PR #358, opened 2026-05-22
Parallel pytato PyOpenCL array context

majosm:
Adds `PytatoParallelPyOpenCLArrayContext`, which is like `PytatoPyOpenCLArrayContext` but parallelizes over the device. Based on @kaushikcfd's work in https://github.com/inducer/arraycontext/pull/216. This version generalizes that code to work with fused loops.

The basic procedure is:
1. Collect instructions that share inames into sets.
2. Find the "outer" inames (i.e., inames shared between all instructions) in each of those sets.
3. Parallelize over those.

cc @lukeolson

⚠️ Currently WIP. TODO:
- [x] Figure out why some tests are failing (here and in https://github.com/inducer/loopy/pull/1010)
- [ ] Re-run profiling on Tuolumne to make sure I haven't messed anything up
- [ ] Merge loopy deps
- [ ] Add docs

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -9266,6 +9266,16 @@
                 }
             }
         ],
+        "./arraycontext/impl/pytato/parallelize.py": [
+            {
+                "code": "reportUnknownVariableType",
+                "range": {
+                    "startColumn": 51,
+                    "endColumn": 79,
+                    "lineCount": 1
+                }
+            }
+        ],
         "./arraycontext/loopy.py": [
             {
                 "code": "reportUnknownMemberType",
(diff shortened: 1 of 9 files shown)


majosm:
@inducer FWIW, I found a workaround for the Intel issues. Seems like the failures are due to buffer overflow, possibly from bad handling of the partially-filled workgroups at the tail ends of loops. Adding some padding at the end via custom allocator makes the failures go away (I tried it on several Intel CL versions, which were previously failing on different tests for each version). It's not an ideal solution, but it avoids having to fiddle with versions or skip tests at least. Thoughts?

PR #360, opened 2026-06-04
Gmres

inducer:
Since I want access to a solve in grudge, I figured moving GMRES from pytential would be the easiest. Once this is merged, I'll remove the one in pytential (with deprecation blah). This one isn't totally compatible, fwiw: The pytential one had some actx autodiscovery "stuff" that I'll probably have to re-add.

This also brings sumpy's `build_matrix` here.


arraycontext/linalg/solve.py
```diff
@@ -0,0 +1,456 @@
...
+# Modified Python port of ./Apps/Acoustics/root/matlab/gmres_restart.m
+# from hellskitchen.
+# Necessary because SciPy gmres is not reentrant and thus does
+# not allow recursive solves.
+
+
+class InnerProduct(Protocol, Generic[T]):
+    """A :class:`~typing.Protocol` for the inner product used by :func:`gmres`."""
+
+    def __call__(self, a: T, b: T) -> T: ...
```
alexfikl, line 74:
```suggestion
    def __call__(self, a: T, b: T) -> float: ...
```
?

arraycontext/linalg/solve.py
```diff
@@ -0,0 +1,456 @@
...
+    """
+
+    if type(x) is not type(y):
+        raise TypeError("'structured_vdot' entries have different types: "
+                f"{type(x).__name__} and {type(y).__name__}")
+
+    from numbers import Number
+    if (isinstance(x, Number)
+            or (isinstance(x, np.ndarray) and x.dtype.char != "O")):
+        return np.vdot(x, y)
```
alexfikl, line 108: Seeing as we have a `NumpyArrayContext` now, does this make sense (the `isinstance(x, ndarray)` part)? Not quite sure where this was used in `pytential` either..

arraycontext/linalg/solve.py
```diff
@@ -0,0 +1,451 @@
...
+    :arg inner_product: a callable with an interface compatible with
+        :func:`numpy.vdot` that returns a host scalar.
+    :arg maxiter: the maximum number of iterations permitted.
+    :arg hard_failure: if *True*, raise :exc:`GMRESError` in case of failure.
+    :arg stall_iterations: number of iterations with residual decrease
+        below *no_progress_factor* indicates stall. Set to ``0`` to disable
+        stall detection.
+    """
+    if inner_product is None:
+        if actx is None:
```
alexfikl, line 386: We can still try to find this from `rhs`? But I guess the caller can do that too, so not a a big issue.

arraycontext/linalg/solve.py
```diff
@@ -0,0 +1,451 @@
...
+            require_monotonicity=require_monotonicity)
+
+
+# }}}
+
+
+def build_matrix(
+            op: CallableOperator[onp.Array1D] | HasMatVec[onp.Array1D],
+            dtype: np.dtype[Any] | None = None,
+            shape: tuple[int, int] | None = None
```
alexfikl, line 407: This looks quite "not arraycontext"-y? i.e. I would expect it to also take an `CallableOperator[ArrayOrContainerT]`. Probably not worth it..

arraycontext/linalg/solve.py
```diff
@@ -0,0 +1,451 @@
...
+    import numpy.linalg as la
+
+    mat = build_matrix(op)
+
+    print(f"condition number: {la.cond(mat)}")
+    if show_spectrum:
+        ev = la.eigvals(mat)
+        import matplotlib.pyplot as pt
+        pt.plot(ev.real, ev.imag, "o")
+        pt.show()
```
alexfikl, line 447: I would remove this?

alexfikl (commented):
Left a few comments, but overall this looks like a good place for it to me.

arraycontext/linalg/solve.py
inducer: I'm trying to move away from actx auto-discovery. For some arrays (e.g. numpy), it can't be done, so I think it's best not to rely on it.

arraycontext/linalg/solve.py
inducer: Not really: GPU array contexts return reduction results as device scalars, i.e. arrays. And numpy is forgiving enough that it sort-of fits.

arraycontext/linalg/solve.py
inducer: I agree, but we can add it when we need it.

arraycontext/linalg/solve.py
inducer: It does make sense to me. :grin: This gets used on object arrays that look like `[DOFArray, scalar]`, e.g. when there are scalar degrees of freedom floating around somewhere.

arraycontext/linalg/solve.py
inducer: Don't know. It's not meant for serious use because of the inherent inefficiency, so it doesn't matter? And six times out of ten, the reason I'm reaching for this is because the iterative solve died because of some issue in the spectrum. Compromise: `_show_spectrum`?

arraycontext/linalg/solve.py
alexfikl: Then this doesn't match what `structured_vdot` is doing? That does a `to_numpy` and takes the item inside. Also things like `residual_norms: Sequence[float]` would need to be updated to hold device scalars?

arraycontext/linalg/solve.py
alexfikl: What I wasn't sure about here is.. if we're using e.g. a `PyOpenCLArrayContext`, how would a non-object numpy array sneak in here that we need to guard against it?

arraycontext/linalg/solve.py
alexfikl: Hm, fair enough, if it's meant more for debugging, it's fine as is. Maybe these deserve a comment or docstring warning me about what I'm getting into? :grin:

PR #365, opened 2026-07-01
Update some deprecated functions

alexfikl:


alexfikl:
Hm, can't say I have any idea about what `conda` suddenly doesn't like. 

A quick diff with the CI from last weekend shows that pretty much nothing changed. The only difference I can see is from https://github.com/conda-forge/conda-feedstock/pull/300..

EDIT: Crisis averted, it's been fixed :grin:

inducer:
Thx!

